PR #55, opened 2022-10-23
feat: hide !rpc and !validate from chat

diznq:


ccomrade (approved):
It's a dirty hack for broken servers, but if we first check `sourceId == targetId`, it's fine for me.

PR #93, opened 2023-05-06
Add DrawTools

griseraner:
Is this acceptable? :D

### DIFF
--- a/Code/CryMP/Client/Client.cpp
+++ b/Code/CryMP/Client/Client.cpp
@@ -28,6 +28,7 @@
 #include "EngineCache.h"
 #include "ParticleManager.h"
 #include "FlashFileHooks.h"
+#include "DrawTools.h"
 
 #include "config.h"
 
@@ -236,6 +237,7 @@ void Client::Init(IGameFramework *pGameFramework)
 	m_pEngineCache       = std::make_unique<EngineCache>();
 	m_pParticleManager   = std::make_unique<ParticleManager>();
 	m_pFlashFileHooks    = std::make_unique<FlashFileHooks>();
+	m_pDrawTools		 = std::make_unique<DrawTools>();
 
 	// prepare Lua scripts
 	m_scriptMain         = WinAPI::GetDataResource(nullptr, RESOURCE_SCRIPT_MAIN);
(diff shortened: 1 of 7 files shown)


Code/CryMP/Client/Client.cpp
```diff
@@ -236,6 +237,7 @@ void Client::Init(IGameFramework *pGameFramework)
 	m_pEngineCache       = std::make_unique<EngineCache>();
 	m_pParticleManager   = std::make_unique<ParticleManager>();
 	m_pFlashFileHooks    = std::make_unique<FlashFileHooks>();
+	m_pDrawTools		 = std::make_unique<DrawTools>();
```
ccomrade, line 240: Tabs instead of spaces. Breaks alignment when different tab width is used.

Code/CryMP/Client/DrawTools.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+	++m_idGenerator;
+	image.id = m_idGenerator;
+
+	m_ImageData.emplace_back(std::move(image));
+
+	return m_idGenerator;
+}
+
+void DrawTools::RemoveTextById(int id)
```
ccomrade, line 135: This function can remove both texts and images, so its name should be something like `RemoveTextOrImageById`.

Code/CryMP/Client/DrawTools.cpp
```diff
@@ -0,0 +1,167 @@
...
+int DrawTools::Add(Image image, const char *texturePath)
+{
+	if (!texturePath)
+		return -1;
+
+	ITexture* pTexture = gEnv->pRenderer->EF_LoadTexture(texturePath, FT_FROMIMAGE, eTT_2D);
+	if (!pTexture)
+		return -1;
+
+	return Add(image, pTexture->GetTextureID());
```
ccomrade, line 102: If this fails, a texture is leaked. `Add` always succeeds, but only for now.

ccomrade (approved):
Looks good

Code/CryMP/Client/DrawTools.cpp
griseraner: I know I thought the same, but image is a Texture so I thought let's keep it short :D

PR #100, opened 2023-08-05
Embedded entity XMLs

ccomrade:
Move data from all entity XMLs directly into C++ code. This allows us to easily fix entity data and prevents users from cheating by modifying these XML files. It also slightly speeds up game startup.

- [x] `Scripts/Entities/Items/XML/Ammo` - loaded by `WeaponSystem` (CryGame)
- [x] `Scripts/Entities/Items/XML` - loaded by `ItemSystem` (CryAction)
- [x] `Scripts/Entities/Vehicles` - loaded by `VehicleSystem` (CryAction)

CryAction source code is not available, so reverse-engineered `ItemSystem` and `VehicleSystem` are added.

This is our first step into CryAction.


ccomrade:
Rebased on top of current master.

griseraner (approved):
Great stuff

PR #104, opened 2023-09-03
Vehicle damage sounds

griseraner:


ccomrade (approved):
Looks good, except that crazy part with `SetSoundParam`.

PR #111, opened 2023-10-29
Improve CFlashMenuObject::UpdateNetwork

ccomrade:
~~Remove `CFlashMenuObject::UpdateNetwork` function, which calls `setNetwork` Flash function on frame. Does it actually do something other than showing no connection symbol in multiplayer lobby? @griseraner~~

Update: Use our own `ServerBrowser::LastRequestSucceeded` instead of problematic `INetwork::HasNetworkConnectivity`.

This fixes no connection symbol always shown in multiplayer lobby under Linux :penguin:, where `INetwork::HasNetworkConnectivity` always returns false when there's no recent network channel activity. In this case, it does some questionable checks of network adapters via Windows API. Definitely not something we want in main thread.

`INetwork::HasNetworkConnectivity` is also called on frame in `CHUD::OnPostUpdate`, but that one should be fine because it's called only in-game and only when `gEnv->bMultiplayer` is true. Therefore, network channel activity is there and it returns quickly without calling potentially blocking Windows API functions.

Thanks to Furyaner for pointing out that `INetwork::HasNetworkConnectivity` is the culprit here. :slightly_smiling_face:

### DIFF
--- a/Code/CryGame/Menus/FlashMenuObject.cpp
+++ b/Code/CryGame/Menus/FlashMenuObject.cpp
@@ -4016,7 +4016,7 @@ void CFlashMenuObject::UpdateNetwork(float fDeltaTime)
 		return;
 	}
 
-	if (!m_multiplayerMenu->IsInLobby() && !m_multiplayerMenu->IsInLogin())
+	if (!m_multiplayerMenu->IsInLobby() && !m_multiplayerMenu->IsInLogin() && !gEnv->bMultiplayer)
 	{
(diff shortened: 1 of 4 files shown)


ccomrade:
Now multiplayer lobby shows no connection symbol when server browser request fails. This is more reliable than network adapter checks in `INetwork::HasNetworkConnectivity` because connected network adapter does not mean that we have internet connection or that master server is available.

griseraner (approved):

PR #124, opened 2023-11-30
Add repair smoothing

griseraner:


Code/CryGame/HUD/HUDInterfaceEffects.cpp
```diff
@@ -1593,20 +1622,34 @@ void CHUD::ShowProgress(int progress, bool init /* = false */, int posX /* = 0 *
 
 		m_iProgressBarX = posX;
 		m_iProgressBarY = posY;
-		m_sProgressBarText = string(text);
+		if (text.data())
+		{
+			m_sProgressBarText = (string)text.data();
+		}
```
ccomrade, line 1628: This is problematic because it doesn't clear `m_sProgressBarText` when `text` is completely empty. Also, `.empty()` should be used in conditions.

If `m_sProgressBarText` is `std::string`, then you can do just:
```suggestion
		m_sProgressBarText = text;
```

Code/CryGame/HUD/HUDInterfaceEffects.cpp
```diff
@@ -1553,38 +1554,66 @@ void CHUD::SetSubtitleMode(HUDSubtitleMode mode)
...
 			else
 				m_animProgress.Load("Libs/UI/HUD_ProgressBar.gfx", eFD_Center, eFAF_Default);
+			
 			m_iProgressBar = 0;
+			m_progressBarSmoothed = -1.0f;
 		}
 
 		pAnim->Invoke("showProgressBar", true);
-		const wchar_t* localizedText = (text)?LocalizeWithParams(text, true):L"";
+		const wchar_t* localizedText = (text.data() ? LocalizeWithParams(text.data(), true) : L"");
```
ccomrade, line 1616:
```suggestion
		const wchar_t* localizedText = text.empty() ? L"" : LocalizeWithParams(text.data(), true);
```

ccomrade (commented):
Looks good overall.

PR #127, opened 2023-12-28
Fix movement bug occuring after unfreeze

griseraner:


ccomrade:
The actual fix is in https://github.com/crymp-net/client-server/commit/654943f7944f189c48e0e3df1079598445253f1e. This PR is merely a cleanup.

ccomrade (approved):

PR #129, opened 2024-01-19
CPPAPI.AddLocalizedLabel: Use existing label for unspecified values

ccomrade:
Instead of having to specify all attributes of a localized label, you can now only set those you want to change and the rest will be taken from the existing label.


ccomrade:
Tested and works.

PR #132, opened 2024-02-01
Block StopEngine request

griseraner:
This should ideally be fixed in CryAction, I suspect StopEngine is called with a delay sometimes, causing it to be called after StartEngine -> silent vehicle with driver


ccomrade (approved):
Looks good except those TABs inside `CryLogAlways` :smile:

PR #134, opened 2024-02-16
Simplify cvar and command declaration to a single line of each

LegendaryGuard:
It's an improvement in development efficiency by simplifying the declaration of cvars and commands.
Previously, it was necessary to write 4 lines of code and modify 2 different files to declare and unregister a cvar, 2 lines of code to declare and unregister a command.

With this modification, the declaration can be done in a single line of code in a single file, significantly reducing the time and effort required.
Not only simplifies the development process but also enhances code readability by reducing redundancy.
This improvement will contribute to faster and more efficient development in the project.

Based on OpenJK, thanks to Razor.


ccomrade (commented):
This is a good idea. I like it. Although all those new macros look a bit scary. What about going even further?

`CGame::UnregisterConsoleCommands` is actually never called, making it a dead code, so it can be removed completely. That means game console commands will be listed only in `CGame::RegisterConsoleCommands`.

About cvars, well, these are more complicated. We can do something like:

```cpp
struct SCVars
{
	IConsole* pConsole = gEnv->pConsole;

	float cl_fov;
	ICVar* p_cl_fov = pConsole->Register("cl_fov", &cl_fov, 60.0f, 0, "Field of view");
	// ...
};
```

Registering all game cvars when constructing `SCVars`. Maybe `m_pCVars = new SCVars();` will need to be moved from `CGame::CGame` to `CGame::Init`, where `SCVars::InitCVars` is called via `CGame::RegisterConsoleVars`. Both these functions can now go away. Even `CGame::RegisterConsoleCommands` isn't necessary anymore because console commands can simply be registered in `SCVars::SCVars`.

I would again do only cvars registration. They are never unregistered anyway and if they were, it would only slow down game shutdown and probably even cause some problems (potential accesses to already unregistered cvars). One never knows in this messy game. :smile:

Now we still have the issue that names of most cvars need to be specified twice like in the `cl_fov` example above. However, this is something a macro can solve in a simple and clean way.

What do you think?

PR #137, opened 2024-03-01
Firemode view transition

griseraner:


Code/CryGame/Actors/Player/Player.cpp
```diff
@@ -5078,6 +5078,11 @@ void CPlayer::PlayAction(const char* action, const char* extension, bool looping
 	if (!m_pAnimatedCharacter)
 		return;
 
+	if (strcmp(action, "use_lockpick") && !strcmp(extension, "lockpick")) //CryMP: A bit hacky, but use lockpick pose only when actually using it
```
ccomrade, line 5081: Is this really correct? Maybe missing `!`?
```suggestion
	if (!strcmp(action, "use_lockpick") && !strcmp(extension, "lockpick")) //CryMP: A bit hacky, but use lockpick pose only when actually using it
```

ccomrade (approved):
Looks good

Code/CryGame/Actors/Player/Player.cpp
griseraner: Yes, so it skips using "broken back" pose when just holding the weapon 🤪

PR #143, opened 2024-04-28
CPPAPI.CreateKeyFunction

griseraner:


Code/CryMP/Client/Client.cpp
```diff
@@ -397,6 +453,10 @@ void Client::ClearKeyBinds()
 {
 	const auto createdInGame = [](KeyBind& bind) -> bool
 	{
+		if (bind.function && bind.createdInGame)
+		{
+			gEnv->pScriptSystem->ReleaseFunc(bind.function);
```
ccomrade, line 458: `createdInGame` shouldn't have any side effects. It's supposed to only check whether a key bind should be removed or not. I think `SmartScriptFunction` should be used instead of raw `HSCRIPTFUNCTION`, so `ReleaseFunc` is called automatically when needed.

Code/CryMP/Client/Client.h
```diff
@@ -67,6 +67,7 @@ class Client : public IGameFrameworkListener, public ILevelSystemListener, publi
 	{
 		std::string key;
 		std::string command;
+		HSCRIPTFUNCTION function = nullptr;
```
ccomrade, line 70:
```suggestion
		SmartScriptFunction function;
```
And then remove all `ReleaseFunc` calls.

Code/CryMP/Client/Client.cpp
```diff
@@ -382,13 +382,69 @@ void Client::AddKeyBind(const std::string_view& key, const std::string_view& com
...
+{
+	for (KeyBind& bind : m_keyBinds)
+	{
+		if (bind.key == key)
+		{
+			if (bind.function)
+			{
+				gEnv->pScriptSystem->ReleaseFunc(bind.function);
+			}
+			bind.function = function;
```
ccomrade, line 395:
```suggestion
			bind.function.Init(gEnv->pScriptSystem, function);
```

Code/CryMP/Client/Client.cpp
```diff
@@ -382,13 +382,69 @@ void Client::AddKeyBind(const std::string_view& key, const std::string_view& com
...
+			return;
+		}
+	}
+
+	KeyBind& bind = m_keyBinds.emplace_back();
+	bind.key = key;
+	bind.function = function;
```
ccomrade, line 402:
```suggestion
	bind.function.Init(gEnv->pScriptSystem, function);
```

ccomrade (approved):

PR #183, opened 2024-08-16
Add System.GetEntityScriptFilePath script function

shortcut0:
get lua script file path from entities


ccomrade (approved):
Looks good. Thanks.

PR #188, opened 2024-08-19
Add fix for invisible weapon after sv_restart

griseraner:


griseraner:
Possibly fixes the bug with invisible weapons in multiplayer?

ccomrade (approved):
This might have some unwanted side effects.

PR #194, opened 2024-08-25
191 change friendly cross color to blue

griseraner:
Fix random crosshair displays


Code/CryGame/Items/Weapons/VehicleWeapon.cpp
```diff
@@ -305,9 +305,12 @@ void CVehicleWeapon::UpdateFPView(float frameTime)
 //---------------------------------------------------------------------------
 void CVehicleWeapon::CheckForFriendlyAI(float frameTime)
 {
+	if (gEnv->bMultiplayer)
+		return;
```
ccomrade, line 309: What about multiplayer AI in the future?

ccomrade (approved):

Code/CryGame/Items/Weapons/VehicleWeapon.cpp
griseraner: Possibly, but this is just original code and not needed in mp for now :D

PR #197, opened 2024-08-27
Add more System and Weapon script bindings

shortcut0:
automation ftw :3

### DIFF
--- a/Code/CryScriptSystem/ScriptBindings/ScriptBind_System.cpp
+++ b/Code/CryScriptSystem/ScriptBindings/ScriptBind_System.cpp
@@ -1840,11 +1840,11 @@ int ScriptBind_System::GetItemClasses(IFunctionHandler * pH)
 
 	for (pEntityRegistry->IteratorMoveFirst(); pClass = pEntityRegistry->IteratorNext();)
 	{
-		if (pClass) {
-			if (pItemSystem->IsItemClass(pClass->GetName()))
-				pObj->SetAt(k++, pClass->GetName());
-
+		if (pItemSystem->IsItemClass(pClass->GetName()))
+		{
+			pObj->SetAt(k++, pClass->GetName());
 		}
+
 	}
 
 	return pH->EndFunction(*pObj);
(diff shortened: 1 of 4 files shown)


Code/CryScriptSystem/ScriptBindings/ScriptBind_System.cpp
```diff
@@ -1823,4 +1827,46 @@ int ScriptBind_System::GetEntityScriptFilePath(IFunctionHandler* pH, const char*
...
+	IEntityClassRegistry* pEntityRegistry = gEnv->pEntitySystem->GetClassRegistry();
+	IEntityClass* pClass;
+
+	for (pEntityRegistry->IteratorMoveFirst(); pClass = pEntityRegistry->IteratorNext();)
+	{
+		if (pClass) {
+			if (pItemSystem->IsItemClass(pClass->GetName()))
+				pObj->SetAt(k++, pClass->GetName());
+
+		}
```
ccomrade, line 1847:
```suggestion
		if (pItemSystem->IsItemClass(pClass->GetName()))
		{
			pObj->SetAt(k++, pClass->GetName());
		}
```

When `pClass` is null, the end has been reached, so the loop body is not called again.

Code/CryScriptSystem/ScriptBindings/ScriptBind_System.cpp
```diff
@@ -1823,4 +1827,46 @@ int ScriptBind_System::GetEntityScriptFilePath(IFunctionHandler* pH, const char*
...
+	IEntityClassRegistry* pEntityRegistry = gEnv->pEntitySystem->GetClassRegistry();
+	IEntityClass* pClass;
+
+	for (pEntityRegistry->IteratorMoveFirst(); pClass = pEntityRegistry->IteratorNext();)
+	{
+		if (pClass) {
+			if (pVehicleSystem->IsVehicleClass(pClass->GetName()))
+				pObj->SetAt(k++, pClass->GetName());
+
+		}
```
ccomrade, line 1868:
```suggestion
		if (pVehicleSystem->IsVehicleClass(pClass->GetName()))
		{
			pObj->SetAt(k++, pClass->GetName());
		}
```

When `pClass` is null, the end has been reached, so the loop body is not called again.

ccomrade (changes requested):

ccomrade (approved):
Looks good. Thanks.

PR #204, opened 2024-09-07
160 add easy way to change character model

griseraner:


ccomrade (approved):
Very nice! I found only a few minor things that can be improved.

PR #221, opened 2024-09-30
Soundmoods supported in FP/TP spectating + bugfix

griseraner:


Code/CryGame/Actors/Player/NanoSuit.cpp
```diff
@@ -1058,7 +1056,7 @@ void CNanoSuit::DeactivateSuit(float time)
 
 void CNanoSuit::PlaySound(ENanoSound sound, float param, bool stopSound)
 {
-	if (!gEnv->pSoundSystem || !m_pOwner || !m_active)
+	if (!gEnv->pSoundSystem || !m_pOwner || !m_active || gEnv->pSystem->IsDedicated())
```
ccomrade, line 1059: `gEnv->pSoundSystem` should be null when running a dedicated server, right?

ccomrade (approved):
Nice refactoring

PR #223, opened 2024-10-02
Fix crash when spawning US_trolley

ccomrade:
The original `Vehicle::InitDamages` must be called always during vehicle initialization.


griseraner:
Nice fix

PR #235, opened 2024-11-02
Add startup time measurement

ccomrade:
Now we know if startup time is getting better or worse.


ccomrade:
image

griseraner:
Cool feature, but maybe change color to white or green Crymp msg, so it looks like this log ?
-----------------------------------------------------
*LOADING: Level Multiplayer/PS/Refinery loading time: 12.18 seconds
-----------------------------------------------------

griseraner (approved):

ccomrade:
image

PR #255, opened 2024-11-21
Fix missing item in hand after leaving spawntruck + potential AI bug

griseraner:


ccomrade (approved):
Nice catch. `passenger` was always nil.
```lua
local passenger = System.GetEntity(passengerId); --CryMP: was missing
```

PR #263, opened 2024-12-29
Add weather system using synchronized server storage

diznq:
## Intro

With this feature, servers are able to set the following:
- TOD variables
- Wind speed and direction
- A synchronized list of active particle effect (or particle effects to remove if active, i.e. default rain on Refinery)
- Global layer settings for snow and rain

## API

To enable the feature, server must set `g_gameRules.game:SetSynchedGlobalValue(2000, true)`
To disable it, server can either clear the value or set it to `false`.

### Time of Day settings

To set TOD variables, server can set variable IDs of  `2100 + tod_id`, i.e. `2110` would be global fog density, given global fog density = TOD variable 10. 

The specified value must be a string in the following format:

- `r1234.56` to set raw value
- `i0.123` or just `0.123` to set interpolated value between minimum and maximum thresholds for this TOD variable
- `v123.45,123.45,123.45` to set a vector value (such as for colors)
- empty string to reset to default

The values must be string, as synchronized storage is not really good at synchronizing raw floats.

### Wind settings

To set the wind value, server can set variable `2001` in the following format:
- `v123.45,123.45,123.45` to set a vector value
- `123.45` to set a coefficient of original value (`original_value * coefficient`)
- empty string to reset to default

### Particle effects synchronization

To synchronize particle effects, server must set a space delimited list of particle effect names as variable ID `2002`

The format is as follows:
- `particle.effect.name` to add a particle effect
- `-particle.effect.name` to remove a particle effect
- empty string to reset to default

Particle string can contain additional information about position, direction, scale and number of instances in the following format `particle.effect.name;key1=value1;key2=value2;key3=value3`

Valid key value pairs for now are:
- `x=123.45` - scale (default: 1)
- `p=123.45,123.45,123.45` - position (default: 512,512,64)
- `d=0.1,0.2,0.3` - direction (default: 0,0,1)
- `i=123` - number of instances (default: 0)

Position's Z coordinate is always aligned to terrain elevation at given location.

### Environment layers synchronization

To synchronize environment layers, server must set a space delimited list of layer names as variable ID `2003`

Valid values for now are: `frozen`, `wet`.

ScreenShot0044
ScreenShot0047


griseraner:
Looks good!
`g_gameRules.game:SetSynchedGlobalValue(2000, "1")` a bool be used for this btw, synched storage supports that

diznq:
@griseraner 
I tried using bool, but it had same issue as floats, setting it to `false` didn't synchronize back to client. Similar issue with floats when you set them to `1.0` or `< 0.0`, it wouldn't get synchronized. That's why I stuck with string, because that one is guaranteed to synchronize always

diznq:
I added one more feature into this: global layer settings. Enabling server to set global frozen or wet environment, which will enable frozen/wet weapon effects (a feature that was already in the game, but not being used) and also applies `MTL_LAYER_FROZEN` / `MTL_LAYER_WET` on all static objects and vegetation.

I think it looks fancy when all things are combined:

ScreenShot0040
ScreenShot0039

```
g_gameRules.game:SetSynchedGlobalValue(2000, "1")
g_gameRules.game:SetSynchedGlobalValue(2001, "4")
g_gameRules.game:SetSynchedGlobalValue(2002, "snow.snow.snow")
g_gameRules.game:SetSynchedGlobalValue(2003, "frozen")
g_gameRules.game:SetSynchedGlobalValue(2109, "0.05")
g_gameRules.game:SetSynchedGlobalValue(2110, "0.05")
```

griseraner:
This is fantastic! Did you manage to apply frozen layer on vegetations and terrain too? It only worked for static objects with the GetObjectsByType for me, but maybe I did something wrong :D 

Btw, SetSynchedGlobalValue(2000, true/false) and low number values works for me, you can dump those values with `dumpss` command. But maybe because I added the rewritten SynchedStorage from client to server.

diznq:
@griseraner Terrain I had to fake using `e_terrain_texture_debug 1` and vegetation seems kinda problematic, because GetObjectsByType returns zero objects there although it should return list of all vegetaton render nodes 😖 I guess that must be fixed somehow :D

Code/CryMP/Client/WeatherSystem.cpp
```diff
@@ -0,0 +1,438 @@
...
+	
+	currently_enabled = false;
+	original_weather_values.clear();
+	active_values.clear();
+	original_wind.reset();
+	active_effects.clear();
+}
+
+void CWeatherSystem::Update(float frameTime) {
+	CServerSynchedStorage *pSSS = g_pGame->GetServerSynchedStorage();
```
griseraner, line 82:
```suggestion
	CSynchedStorage *pSSS = g_pGame->GetSynchedStorage();
```
Client can't read from server synched storage, probably why changes didnt work including bool/ints, guess you tested on LAN server? Even floats you can use for this, as long as you dont need huge precision,

Code/CryMP/Client/WeatherSystem.cpp
```diff
@@ -0,0 +1,438 @@
...
+	if (!pSSS) {
+		return;
+	}
+
+	m_time += frameTime;
+
+	string b;
+	if (!pSSS->GetGlobalValue(WEATHER_NAMESPACE + WEATHER_ENABLED, b) || b.length() == 0 || b == "0") {
```
griseraner, line 90:
```suggestion
	bool enabled = false;
        pSS->GetGlobalValue(WEATHER_NAMESPACE + WEATHER_ENABLED, enabled);
        if (!enabled) {
```

diznq: Updated this one, with CSynchedStorage it works, but I would still leave those strings in place instead of float as it allows for greater flexibility. Some TOD fields are floats, some are vectors, also having option to use either raw value `r123.45` ot interpolated `123.45` makes things much easier in case you don't remember correct `min`, `max` values for each TOD 😄

griseraner: Sounds good 👍 
GetObjectsByType(eERType_Vegetation) returns several thousand rendernodes for me, on mesa its 100k :D it doesn't for you? 
You can specify eERType_RoadObject_NEW (not the other one) and it will return roads as well

griseraner: The athmosphere is really cool, with the snow effects. Especially at night :)

diznq: Well, it works now, all of sudden, I guess I just had old code (created this branch month ago, didn't sync with master :D), but it's weird, still with MTL_LAYER_FROZEN vegetation isn't frozen. I even tried using DYNFROZEN, but it didn't help.
Gotta find out how it works in Sandbox editor :D As there it clearly works when you enable Layer_frozen for vegetation

diznq: I made entities it should freeze / wet an global array (static_entities) so feel free to experiment with that on what everything to freeze, as of not there is only three types (brushes, vegeatation, ropes) 😄

griseraner: Yea. Just dumped mesa stats. Actually its almost 300k vegetations on Mesa.. :D
Freezing everything would be super cool literally ;)

EERType: Unknown, Count: 685
EERType: Brush, Count: 6729
EERType: Vegetation, Count: 290768
EERType: Light, Count: 66
EERType: Cloud, Count: 118
EERType: FogVolume, Count: 3
EERType: Decal, Count: 1060
EERType: ParticleEmitter, Count: 77
EERType: WaterVolume, Count: 2
EERType: RoadObject_NEW, Count: 944
EERType: DistanceCloud, Count: 119

diznq: I guess I figured it out! :D
ScreenShot0048

Just gotta fix some issues to make it more pretty :D

griseraner: Omg 👯

diznq: Pushed it, feel free to try out. Trick is that vegetation is actually an instanced stat object, so this code just naively iterates all stat instance group IDs from 0 to 65536 (sadly there is no GetGroupIDs, but let's assume it's never like 1 000 000+) and sets material flags on those instead.

diznq:
Pushed one more fix that handles `nicegrass` and `bigpatch` types of vegetation. Usually setting `MAT_LAYER_FROZEN` on these two types ends up being really ugly, so it's preferrable to not set `MAT_LAYER_FROZEN` but instead use `frozen_nicegrass_...cgf` for `bigpatch_..._frozen.cgf` models.

Here is the end result
ScreenShot0050
ScreenShot0051

griseraner:
Nice! Those big trees could have frost layer, or replaced by trees without leafs maybe. Anyway getting better and better :D

diznq:
They do have a frost layer, as every other instance of vegetation. It's just that frost layer in Crysis really is just a shader that dynamicly computes amount of snow / frost on a vertex based on it's normal. The more it's up-facing, the more snow it receives.

You can check this out, I remember reading it long time ago on how they did it: https://www.slideshare.net/slideshow/crysis-nextgen-effects-gdc-2008/25051981#8 Slides 33 to 38 😄

ccomrade:
`GetObjectsByType` is fixed now (#262)

Code/CryMP/Client/WeatherSystem.cpp
```diff
@@ -0,0 +1,569 @@
...
+#include "ScriptBind_CPPAPI.h"
+#include "Client.h"
+#include "ScriptCommands.h"
+#include "ScriptCallbacks.h"
+#include "DrawTools.h"
+#include "CryGame/GameActions.h"
+#include "CryGame/Actors/Actor.h"
+#include "CryGame/GameRules.h"
+
+#include <map>
```
ccomrade, line 26:
```suggestion
#include <array>
#include <map>
#include <set>
#include <string>
#include <string_view>
#include <sstream>
#include <vector>

#include "CryCommon/CrySystem/ISystem.h"
#include "CryCommon/Cry3DEngine/I3DEngine.h"
#include "CryCommon/CryRenderer/IRenderer.h"
#include "CryGame/Game.h"

#include "WeatherSystem.h"
```

ccomrade (approved):
Cool stuff

diznq:
I think it's good to go now, I fixed some stability issues as it was crashing when map reloaded due to previously replaced geometry on the fly. Also fixed the way particle effects are disabled. Now it's lot nicer, rather than immediately stopping snow, it lets snow fall off and then doesn't emit any further snow.

Here is the final comparison :D
ScreenShot0054
ScreenShot0056

If you wanna try it out on your server, you can use these presets:

```lua
	g_gameRules.game:SetSynchedGlobalValue(2000, true)
	g_gameRules.game:SetSynchedGlobalValue(2001, "3")
	g_gameRules.game:SetSynchedGlobalValue(2002, "snow.snow.snow;x=3;i=4")
	g_gameRules.game:SetSynchedGlobalValue(2003, "frozen")

	g_gameRules.game:SetSynchedGlobalValue(2104, "0.03")
	g_gameRules.game:SetSynchedGlobalValue(2108, "v1.0,1.0,1.0")
	g_gameRules.game:SetSynchedGlobalValue(2109, "0.03")
	g_gameRules.game:SetSynchedGlobalValue(2110, "0.02")
	g_gameRules.game:SetSynchedGlobalValue(2115, "0")
	g_gameRules.game:SetSynchedGlobalValue(2140, "")

	ForceSet("e_terrain_texture_debug", "1")
```

PR #264, opened 2024-12-31
Add combat organization using radio

diznq:
## Intro

Crysis kinda lacks a way to organize a combat, with voice chat pretty much dead and default radio messages being too cumbersome.

This feature enhances combat by providing a new way to tag a location where to gather.

## Usage

All the player has to do is to hold middle mouse button for less than 0.25s (just quickly open the nano suit menu without actually using it). After this, a radio message "Request assistance" will be sent to the team with location where player was looking at (using raycasts). 

This way it is clear to team mates where to go. If multiple players do this at once, client sees only the closest tagged point. The tagged point also disappears automatically after 9 seconds (same time it takes a chat message to disappear from chat).

**Requests are ignored if player changes nano-suit mode** to prevent usage by mistake if someone is too quick in using nanosuit.

## Implementation

In order to remain compatible with all servers (with or without SSM) it is not possible to introduce information about location into SvRadioMessageParams RMI request, therefore it is routed as a "special" kind of a chat message that begins with `\n`. It is impossible to write a message that begins with `\n` in normal game anyway + this also "hides" the messages for players without client in worst case (as Flash only draws the first line of text - first line being empty in this case).

Syntax therefore is `\n Command Payload`, similar to chat commands. For this feature it would be
`\nradio message_id` for standard radio over chat or `\nradio message_id pos_x pos_y pos_z` for radio over chat with tagged position.

https://github.com/user-attachments/assets/d80bdd26-d50d-4376-a978-e55acff5c206

### DIFF
--- a/Code/CryCommon/CryAction/IGameRulesSystem.h
+++ b/Code/CryCommon/CryAction/IGameRulesSystem.h
@@ -41,14 +41,7 @@ enum EChatMessageType
 {
 	eChatToTarget=0, // the message is to be sent to the target entity
 	eChatToTeam, // the message is to be sent to the team of the sender
-	eChatToAll, // the message is to be sent to all client connected
-};
-
-// Summary
-//    Opcodes for special chat messages
-enum EChatMessageOpcode {
-	eChatOpcodeInvalid = 0,
-	eChatOpcodeRadio
+	eChatToAll // the message is to be sent to all client connected
 };
 
 // Summary
--- a/Code/CryCommon/CryAction/IGameRulesSystem.h
+++ b/Code/CryCommon/CryAction/IGameRulesSystem.h
@@ -41,7 +41,14 @@ enum EChatMessageType
 {
 	eChatToTarget=0, // the message is to be sent to the target entity
 	eChatToTeam, // the message is to be sent to the team of the sender
-	eChatToAll // the message is to be sent to all client connected
+	eChatToAll, // the message is to be sent to all client connected
+};
+
+// Summary
+//    Opcodes for special chat messages
+enum EChatMessageOpcode {
+	eChatOpcodeInvalid = 0,
+	eChatOpcodeRadio
 };
 
 // Summary
--- a/Code/CryGame/GameCVars.cpp
+++ b/Code/CryGame/GameCVars.cpp
@@ -615,6 +615,7 @@ void SCVars::InitCVars(IConsole* pConsole)
 	pConsole->Register("mp_C4StrengthThrowMult", &mp_C4StrengthThrowMult, 1.0f, OPTIONAL_SYNC, "Strength throw mult for C4s");
 	pConsole->Register("mp_netSerializePhysVel", &mp_netSerializePhysVel, 0, OPTIONAL_SYNC, "Serialize the player physics velocity for more precise speed");
 	pConsole->Register("mp_netSerializeMaxSpeed", &mp_netSerializeMaxSpeed, 9.0f, OPTIONAL_SYNC, "Maximum character speed");
+	pConsole->Register("mp_radioTagging", &mp_radioTagging, 0, OPTIONAL_SYNC, "Enable tagging positions using radio");
 
 	//CryMP CVars (un-synced)
 	pConsole->Register("mp_newSpectator", &mp_newSpectator, 1, VF_NOT_NET_SYNCED, "");
@@ -944,6 +945,7 @@ void SCVars::ReleaseCVars()
 	pConsole->UnregisterVariable("mp_wallJump", true);
 	pConsole->UnregisterVariable("mp_flyMode", true);
 	pConsole->UnregisterVariable("mp_messageCenterColor", true);
+	pConsole->UnregisterVariable("mp_radioTagging", true);
 }
 
 //------------------------------------------------------------------------
--- a/Code/CryGame/GameCVars.h
+++ b/Code/CryGame/GameCVars.h
@@ -469,6 +469,7 @@ struct SCVars
 	float		mp_netAimLerpFactor;
 	float		mp_netAimLerpFactorCrymp;
 	int         mp_explosiveSilhouettes;
+	int         mp_radioTagging;
 
 	SCVars();
 	~SCVars();
--- a/Code/CryGame/GameRules.cpp
+++ b/Code/CryGame/GameRules.cpp
@@ -749,8 +749,46 @@ void CGameRules::OnTextMessage(ETextMessageType type, const char* msg,
 //------------------------------------------------------------------------
 void CGameRules::OnChatMessage(EChatMessageType type, EntityId sourceId, EntityId targetId, const char* msg, bool teamChatOnly)
 {
-	if (sourceId == targetId && (strstr(msg, "!rpc ") == msg || strstr(msg, "!validate ") == msg))
-		return;
+	std::string_view sv{ msg };
+	if (sourceId == targetId) {
+		if (sv.starts_with("!rpc ") || sv.starts_with("!validate")) {
+			return;
+		}
+	} else if (sv.starts_with("\n") && sv.length() >= 2) {
+		bool valid = false;
+		// if message starts with \n, consider it a special broadcast from client rather than a legit chat message
+		// this way we can ensure stuff works everywhere and not only servers where SSM supports it,
+		// also starting the message with \n assures that it stays invisible even for people without client
+		// 
+		// the format for commands is as follows: \n (opcode + '0') Payload, i.e. \n11 for Radio 1
+		std::string payload{ sv.substr(2) };
+		uint8_t opcode = (static_cast<uint8_t>(sv[1]) & 0xFF) - '0';
+		switch (opcode) {
+		case EChatMessageOpcode::eChatOpcodeRadio:
+		{
+			// radio over chat
+			int id;
+			float x, y, z;
+			if (sscanf(payload.c_str(), "%d,%f,%f,%f", &id, &x, &y, &z) == 4) {
+				valid = true;
+				OnRadioMessage(SRadioMessageParams{
+					.id = id,
+					.sourceId = sourceId,
+					.pos = Vec3(x, y, z)
+				});
+			}
+			else if (sscanf(payload.c_str(), "%d", &id) == 1) {
+				valid = true;
+				OnRadioMessage(SRadioMessageParams{
+					.id = id,
+					.sourceId = sourceId,
+				});
+			}
+		}
+		break;
+		}
+		if (valid) return;
+	}
 	//send chat message to hud
 	int teamFaction = 0;
 	if (IActor* pClientActor = gEnv->pGame->GetIGameFramework()->GetClientActor())
@@ -4218,18 +4256,18 @@ void CGameRules::SendRadioMessage(const EntityId sourceId, const int msg)
 	}
 }
 
-void CGameRules::OnRadioMessage(const EntityId sourceId, const int msg)
+void CGameRules::OnRadioMessage(const SRadioMessageParams& params)
 {
 	//CryMP: Mute check
 	IVoiceContext* pVoiceContext = gEnv->pGame->GetIGameFramework()->GetNetContext()->GetVoiceContext();
-	const bool muted = pVoiceContext->IsMuted(m_pGameFramework->GetClientActorId(), sourceId);
+	const bool muted = pVoiceContext->IsMuted(m_pGameFramework->GetClientActorId(), params.sourceId);
 	if (muted)
 	{
 		return;
 	}
 
 	//CryLog("[radio] from: %s message: %d",,msg);
-	m_pRadio->OnRadioMessage(msg, sourceId);
+	m_pRadio->OnRadioMessage(params);
 }
 
 void CGameRules::RadioMessageParams::SerializeWith(TSerialize ser)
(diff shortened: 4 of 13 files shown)


ccomrade (commented):
Well, i like the idea. Also the indicator looks nice.

One problem is that it's easy to trigger by mistake and since there's no cooldown, it's trivial to spam with:

image

It's also yet another thing that abuses chat as its transport. We can certainly do better. What about implementing this server-side? It could be an extension of standard Request Assistance radio message. @griseraner any thoughts on this?

diznq:
I would refrain from doing it server side. I think it should be a global feature supported on all servers, not only on chosen ones. (btw I think we should also move those Circle jump enabling CVars and so on to global synch instead of CVars because of this, but that's for another feature :D)

As for trigerring it by mistake, I don't think it's the case. If someone is too quick when it comes to changing nano suit modes, the radio action is prevented. And does anyone just randomly click middle button to do nothing? I don't think so, only if intentionally and not by mistake, as you demonstrated on screenshot :D

diznq:
By the way, you can still do cooldowns on server side here, as in the end, it is server that decides whether message received by SvRequestChatMessage gets propagated further to clients, so server can even treat it as a chat command and deny it 🙄.

ccomrade:
> I would refrain from doing it server side. I think it should be a global feature supported on all servers, not only on chosen ones. Let's stop discriminating other servers, shall we? 😄 There is nothing wrong with using chat for transport here, that way it ensures it works for everyone and everywhere, which should be #&NoBreak;1 priority.

I wish grabbing objects would be available on every server as well. It makes so much sense as it's a standard feature in SP. Maybe we should "just" prepare an official up-to-date SSM finally, so the outdated ones can go. Client gained `-dedicated` recently. It's true that backward compatibility is very important, but trying to make new features work on vanilla servers only brings more headache in my opinion. We already have `mp_crymp` cvar for features incompatible with vanilla servers.

> And in that case if it's intention, there is nothing wrong with getting punished by SSM or something. Even now people just spam radio commands all the time, yesterday I played and it was just "Need backup" 5x in 30 seconds by a player I won't name here 😄. But it didn't get punished because it was normal radio message at that time, not chat message, but annoying nevertheless.

> By the way, you can still do cooldowns on server side here, as in the end, it is server that decides whether message received by SvRequestChatMessage gets propagated further to clients, so server can even treat it as a chat command and deny it 🙄.

True, but client should at least make it less likely to happen. It's been a while, but i remember someone complaining about getting kicked from certain servers when buying stuff too fast for example. Which brings us back to the first point that maybe it shouldn't be available on servers that don't explicitly enable it.

diznq:
> True, but client should at least make it less likely to happen. It's been a while, but i remember someone complaining about getting kicked from certain servers when buying stuff too fast for example. Which brings us back to the first point that maybe it shouldn't be available on servers that don't explicitly enable it.

There is a `0.25s` timer for it to happen, as long as you hold that middle mouse button for longer than that, no radio action happens. So it really needs to be an intention here, not that some newbie opens nanosuit mode by mistake, looks at all options there and then after 7 seconds closes 🙄. I think it should be a global feature, as it's pretty integral to the game to communicate well. In my personal opinion, this is something that Crysis really lacks. You have no way to tell someone where to go during combat, everything is just so random and without coordination that there is no "strategizing" at all 😄. It's just chaotic. So in this way, consider it an improvement of gameplay experience rather than server specific feature.

> I wish grabbing objects was available on every server as well. It makes so much sense as it's a standard feature in SP. Maybe we should "just" prepare an official up-to-date SSM finally, so the outdated ones can go. Client gained -dedicated recently. It's true that backward compatibility is very important, but trying to make new features work on vanilla servers only brings more headache in my opinion. We already have mp_crymp cvar for features incompatible with vanilla servers.

Well grabbing objects is different, because due to physics, it needs to enabled explicitly, so server knows to synchronize that stuff 😄. But this is different, as from point of server, there is nothing it wouldn't be able to synchronize (yeah, it goes as Chat RMI, but so be it, modifying other RMI would just get everyone kicked out on servers that don't support it).

diznq:
In any case, I also added fix so it doesn't trigger if someone quickly selects "Weapon settings" menu from nanosuit, which in previous case would satisfy condition "nanosuit mode didn't change && event happened under 0.25s" 😄. So now doing it by mistake I think is pretty hard.

diznq:
So I'd say, let's really stop caring about vanilla servers that much. It's not like anyone plays on them much, and I don't even think there is vanilla servers without SSM anyway. So for this feature, let's force it globally, BUT if any server wants, it can easily disable that feature by simply preventing chat messages that begin with `\n`, which is fairly simple to do even in SSM SafeWriting.

```lua
AddFunc(function(type, source, target, message)
    if message:sub(1, 1) == "\n" then
        return false
    end
end, "OnChatMessage")
```

Enabled by default, disabled if server handles it that way.

griseraner:
The idea is good, I like it. 
Although, instead of suitmenu, how about showing radios as icons for every radio commands, then we could remove chat messages completely? I think most of the command could be useful to see on the map. And append coordinates (E4) to the message

diznq:
> Although, instead of suitmenu, how about showing radios as icons for every radio commands, then we could remove chat messages completely?

Well, point of this feature is that it appears where you look at, not where you stand at. I don't think it makes point to point where you stand at.

If I want to warn my team mates that "hey, there is a sniper on that hill up there", information about my position is pretty useless to them. But instead I can look at the hill, look at the sniper and tell my team mates "hey, look here, there is a sniper".

Same when I want to tell my team mates "hey, let's take this path left, not right" when approaching a bunker, I can point to the left path and tag it for them. Simple, quick & easy.

So from this perspective, it is not possible to do it over Radio RMI, as you cannot put that "point where I look at" into it without changing RMI itself. You might way, you can compute that on server side by doing a raycast from player's head and head direction, but that won't work if the player is inside a vehicle. As when player is in the vehicle, even if they look around, the head rotation doesn't change + it's very inaccurate too 😄

griseraner:
I see what you mean, it also bothers me that Sniper Spotted (and Armor vehicles etc) are kinda useless if it shows your location instead of sniper location. Hey how about when using binoculars, you hold down left mouse click -> this opens the radio menu and you can "spawn" the radio command at that location?

diznq:
I am not really fan of binoculars as first of all, nobody ever buys those unless server gives them to you for free at spawn and then using binoculars to tag something that is 10m away is kinda cumbersome. 

Say the second case when you are leaving prototype factory on Mesa with team mates and you want to tell them to take left path instead of right path towards bunker. Would you really use binoculars to zoom on something 10 meters away? Plus it also takes some time for binoculars to "load", which is annoying when combat is fast paced :D

But it might be a good separate feature that you could tag entities for others using binoculars as well 😄

ccomrade:
By the way, it creates mess in server log:

```
CHAT Comrade to Team black:
   
radio 15 1565.550 1786.790 80.062
CHAT Comrade to Team black:
   
radio 15 1562.119 1788.948 79.519
CHAT Comrade to Team black:
   
radio 15 1562.570 1788.072 79.519
CHAT Comrade to Team black:
   
radio 15 1562.759 1786.914 79.519
CHAT Comrade to Team black:
   
radio 15 1562.404 1789.936 79.602
CHAT Comrade to Team black:
   
radio 15 1556.010 1801.038 79.204
CHAT Comrade to Team black:
   
radio 15 1552.191 1800.550 79.204
CHAT Comrade to Team black:
   
radio 15 1547.131 1799.195 79.204
CHAT Comrade to Team black:
   
radio 15 1541.323 1795.249 79.204
CHAT Comrade to Team black:
   
radio 15 1536.684 1791.002 79.204
CHAT Comrade to Team black:
   
radio 15 1539.174 1806.987 80.432
CHAT Comrade to Team black:
   
radio 15 1560.999 1818.857 80.209
```

Also, some legacy SSMs might have problems with newlines in chat messages.

What about simply sending an actual `!radio` chat command instead? We can have a cvar for this, so it gets enabled only on servers that implement `!radio` and explicitly request this feature.

diznq:
> By the way, it creates mess in server log

It's this nit picking that keeps this client out of progress for years

diznq:
Maybe one question to ask oneself regarding this project is: Who comes first - player or server?

And if the answer is player, why is all of this server-centric nit-picking required?

ccomrade:
So let's just force enable every random feature everywhere because "who cares", right? Server owners will love us:

image

All i'm asking for is a feature toggle (cvar) for this and to keep it disabled by default. That's it. We need a feature toggle anyway in case some problem is discovered later on. Just to be on the safe side.

diznq:
Why do you care more about server owners than players?

ccomrade:
Even the extended radio menu isn't enabled by default...

diznq:
> I care about both. Players get new features only on servers that support these new features, so certain quality is assured

You know very well, that this is just alibism for vendor locking it on 7Ox as majority of server owners are unable to implement these. Just like any other from thoes `mp_...` CVars.

> Imagine how annoying it would be if something broke on your server just because the client gained a new feature.

Please show me a server that gets broken by a chat message.

diznq:
Also this is client-to-client feature, server shouldn't even care. It's just a shorthand for player manually typing to chat "hey, look there". It makes no sense to condition typing such message by server.

ccomrade:
I'll give you an example. Some time ago, we violated the rule and enabled a new feature by default. It was the third-person view mode. Everything was fine in the beginning. It was available on all servers (it actually still is). But then misterSD started complaining because third-person view is cheating in some of his puzzle maps. See? People spend time designing their servers and then we break it. And that was just third-person view. Here we have something that generates a chat message with a newline, which was never a thing before, on every middle mouse click. Things can go wrong. Not to mention if a server doesn't have a chat flood protection. Anyone can easily spam this. At least scoreboard mute now hides radio messages as well.

By the way, typing "hey look at E4 hill" and simply pressing middle mouse are two pretty different things, aren't they?

diznq:
> But then misterSD started complaining because third-person view is cheating in some of his puzzle maps. 

Third person view obviously indeed is cheating as it allows you to see people behind you to certain extend giving you unfair advantage. It even improves your recoil 😄

> By the way, typing "hey look at E4 hill" and simply pressing middle mouse are two pretty different things, aren't they?

No, you can create a macro `bind x say Hey there is a sniper on hill` which will do exactly same thing. Does that mean I will break server if I press `x` now?

diznq:
See, this is the main issue with CryMP, we care so much about server owners instead of players that there is more servers than players because of that 😄. 

I think servers should bend to client and not client to servers. After all, it's better to have players without servers than servers without players. In your misterSD example, I think he should be grateful he had players there to begin with instead of complaining about third-person view mode.

This would at least push servers to improve, I mean for how long do we for example want to tolerate SSM CryFire not knowing how to do HTTPS and thus allowing HTTP on master server? 🤔😄 See, it's not really a healthy approach to bend to servers, is it?

ccomrade:
> Third person view obviously indeed is cheating as it allows you to see people behind you to certain extend giving you unfair advantage. It even improves your recoil 😄

It doesn't improve recoil at all. Combat is actually a lot harder in TP. Especially because bullets don't go exactly where crosshair is. There's an open issue for that. TP is mostly harmless. Except some corner cases, where it actually causes problems as we discovered. At least players can't bother each other with it. But that's all irrelevant here. The point is that it's bad to enable potentially problematic features by default.

> Ofc, if this was a paid product, this would be a completely different story when it comes to who should bend to whom.

It sounds like we essentially shouldn't care about server owners because they don't pay us. And that they even should be grateful for that or something. That's certainly not the right approach. It's so arrogant. Nobody pays server owners as well. They invest their time and resources for free just like we do. We are in the same boat. Therefore, we should be nice and respect each other. If i was just a server owner and client devs bothered me by pushing features i didn't ask for, i would rather close my server and leave that shit. It would be better than being frustrated with adapting my server to sudden changes in the client.

I think the diverse server portfolio is one of CryMP's strengths. We did quite well in adding new features and maintaining compatibility with legacy servers at the same time. Each server gets what it was designed for. Players can enjoy both worlds. Latest features and a bit of nostalgia. And if something turns out to be a bad thing, they can just switch to another server. It's a lot of effort to keep even one server up to date. @griseraner does incredible job with 7OXICiTY. So the assumption that server owners should keep up with us as we roll out new client features is completely wrong. It's way easier to simply maintain backward compatibility. We don't want to end up with having only 7OXICiTY, do we? The server list would look really sad and there's no way it would magically bring more players or something.

So, please, either add a cvar for this to keep it disabled by default, or at least use the standard radio trigger instead of the middle mouse. Right now, it's very prone to abuse and accidental triggers. I like the idea, but i don't like the current implementation as:
- a developer because of the dirty hack with chat messages, although we can live with it for now i guess
- a server owner because it's enabled by default and there's no easy way to disable it (that chat filtering proposal is just another dirty hack and it creates technical debt on servers that disable this feature)
- a player because:
    - some players will spam it for sure
    - some players, including me, will accidentally trigger it for sure
    - there's no way to disable it locally to avoid accidental triggers

By the way, we haven't even tested it in action yet and we are already about to force enable it everywhere. If something goes wrong, we will have to create another PR to disable it. That's definitely not ideal.

> Anyways, if you have any ideas how to make it work everywhere without asking server owners to implement it, feel free to come up with such implementation. I am giving up :D

Just add a cvar for it, so we can first test it without enabling it everywhere. Server owners can even set these cvars in pure Lua by calling `CPPAPI.FSetCVar` via RPC. I don't get why cvars as feature toggles are suddenly a problem.

ccomrade (approved):
> Your assumption is that server owners know what RPC is and how to use it. There is no documentation on it at all, which is another pain point here I think.

Absolutely. We need an easy-to-use and up-to-date SSM. Foundations are already in place, but there is still a lot of work to be done. Especially in terms of ease of use and documentation.

> Anyways, I added it under mp_radioTagging then with 0 as a default then.

Thanks for that. @griseraner do you think it could be enabled on 7OXICiTY to see how it works in action?

PR #266, opened 2025-01-03
Restrict walljump velocity

diznq:
## Intro

As pointed in an already existing issue, `mp_wallJump` should be an integer with three modes:

- `0` disabled
- `1` enabled default
- `2` enabled restricted

This feature follows on similar premise although with

- `0.0 - 1.0` enabled default with coefficient
- `1.01 +` cap the velocity of player when performing walljump

## Usage

It takes some playing around, but I found ideal value to be somewhere between `17` and `22` range. This way it enables for normal wall jumping, but restricts abuse (such as mouse scroll binding etc).

https://github.com/user-attachments/assets/b5f4a148-bf09-43f6-98f0-8be89c818b9c

### DIFF
--- a/Code/CryGame/Items/Weapons/Fists.cpp
+++ b/Code/CryGame/Items/Weapons/Fists.cpp
@@ -21,6 +21,7 @@ History:
 #include "CryCommon/CryAction/IWorldQuery.h"
 #include "CryCommon/CryAction/IGameTokens.h"
 #include "CryGame/GameActions.h"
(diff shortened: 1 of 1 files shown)


ccomrade (approved):
Thumbs up for keeping `mp_wallJump` as float.

The solution is a bit too complex since we probably only needed to limit the number of impulses per second (there's already `p_max_player_velocity` cvar to limit player velocity globally), but it seems to work.

Thanks.

PR #270, opened 2025-01-11
Integrate advertising system into the client

diznq:
## Advertising system

This feature implements an advertisement system into the client.

It works as follows:

- retrieve a list of advertisements from the advertisement server
- load the PAK with advertisement textures into the game
- replace all billboards and signs in the game with advertisements in randomized order
- accurately track which advertisements were seen by player for how long
- report the statistics about seen ads back to advertising server

ScreenShot0072
ScreenShot0074
ScreenShot0073

### DIFF
--- a/Code/CryMP/Client/Advertising.cpp
+++ b/Code/CryMP/Client/Advertising.cpp
@@ -10,6 +10,8 @@
 #include "CryCommon/CryAction/IMovementController.h"
 #include "CryCommon/CrySystem/ICryPak.h"
 
+#include "CrySystem/RandomGenerator.h"
+
 #include "CryGame/Actors/Actor.h"
 #include "CryGame/Game.h"
 #include "CryGame/GameCVars.h"
@@ -24,7 +26,6 @@
 
 #include <set>
 #include <string>
-#include <random>
 
 #define AD_CYCLE_MAIN 0
 #define AD_CYCLE_VIEW_CHECK 1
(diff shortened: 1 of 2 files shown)


Code/CryMP/Client/Advertising.cpp
```diff
@@ -0,0 +1,592 @@
...
+	"bordercontrol", "bordercheck",
+	"forward", "right" , "left",
+	"sign", "research",
+};
+
+static std::set<std::string> knownAdLayers;
+static std::vector<SAdCue> cues;
+
+static std::random_device rd;
+static std::mt19937 rng(rd());
```
ccomrade, line 62: There's already one global seeded random generator in [CrySystem/RandomGenerator.cpp](https://github.com/crymp-net/client-server/blob/master/Code/CrySystem/RandomGenerator.cpp).

Code/CryMP/Client/Advertising.cpp
```diff
@@ -0,0 +1,592 @@
...
+
+CAdManager::CAdManager() {
+	knownAdLayers.clear();
+	for (auto& k : knownAdLayersOrdered) {
+		knownAdLayers.insert(k);
+	}
+}
+
+void CAdManager::Update(float delta) {
+	if (ADS_ENABLED) {
```
ccomrade, line 80:
```suggestion
void CAdManager::Update(float delta) {
	FUNCTION_PROFILER(gEnv->pSystem, PROFILE_GAME);
	if (ADS_ENABLED) {
```

ccomrade (approved):
I'm not happy that there's AGAIN no way to disable it without editing client code, but whatever. Let's try it in the dev version and see how it goes.

Code/CryMP/Client/Advertising.cpp
ccomrade: `RandomGenerator::GenerateUInt32` -> [Min, Max]
`RandomGenerator::GenerateIndex` -> [Min, Max)

ccomrade:
Nice

ccomrade:
True. It should be fine. And yes, just merge it to master. A new dev release will be created automatically.

PR #272, opened 2025-01-16
Minor client improvements

diznq:
- Add client version into HTTP requests related to CryMP master API calls
- Exclude HWID, timezone, and other client related information in HTTP requests related to advertising
- Do not count ad time if player is idle for more than 5 seconds
- Optmize ad tracking heuristics

### DIFF
--- a/Code/CryMP/Client/Advertising.cpp
+++ b/Code/CryMP/Client/Advertising.cpp
@@ -480,7 +480,6 @@ void CAdManager::CollectViewership(float delta) {
 	// do not collect ads statistics if the player is idle (most likely afk)
 	// as that would be a waste of ad spend
 	if (m_idleTime > 5.0f) {
-		CryLog("Idle for %f seconds", m_idleTime);
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


ccomrade (approved):
Looks good. Those global variables probably should be `CAdManager` member variables. There's always only one instance anyway.

ccomrade:
It's always nice to include minimum stuff, but there's nothing wrong with including CryEngine stuff in a header. I refactor CryCommon headers from time to time as well. There's still a lot to improve, but it's no longer the same garbage as in vanilla Mod SDK. Even that StdAfx nonsense including everything everywhere is gone. If you need `Vec3` or any other CryMath stuff, then:
```cpp
#include "CryCommon/CryMath/Cry_Math.h"
```
Including directly `CryCommon/CryMath/Cry_Vector3.h` without anything else probably won't work right now. CryMath stuff isn't fully cleaned up yet, but we are getting there.

ccomrade:
Well, yes, it's useful, but trading runtime performance for faster compilation isn't the best way.

Modules should be the cleanest solution of include hell. Major compilers and [CMake](https://www.kitware.com/import-std-in-cmake-3-30/) are making good progress there. Eventually, we will be able to do things like:
```cpp
import std;
```
Converting this project into modules will certainly be a lot of fun. And it will further ensure CryMP's technological lead in Crysis modding. :smile:

PR #274, opened 2025-01-25
Fix sky color & fog color in the weather system

diznq:
Sky color and fog color settings were not working in the weather system previously as they are a color spline.

The colors can now be configured as multiplier instead that are applied after TOD interpolations are done.

Here is before and after (notice the shadow colors especially)

ScreenShot0108
ScreenShot0107

ScreenShot0105
ScreenShot0106


ccomrade (approved):
Nice.

By the way, in the next iteration, `TODUpdate` should be moved to its own local class to prevent accidental use of `this`.

PR #275, opened 2025-01-26
Add identifiers into advertising fetch request

diznq:
- Server identifier can be used in future to exclude ads from displaying on a certain server (let's say you don't want your ads to display on your servers)
- Server identifier can be also used to disable ads on server altogether if the ad server recognizes such rule
- Player profile ID identifier can be used to disable ads on a server if the ad server recognizes such rule for given player

### DIFF
--- a/Code/CryMP/Client/Advertising.cpp
+++ b/Code/CryMP/Client/Advertising.cpp
@@ -20,6 +20,7 @@
 
 #include "Client.h"
 #include "FileCache.h"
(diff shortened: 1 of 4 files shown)


ccomrade (approved):
The way to disable ads is a bit meh. I would expect some easy way that servers can tell clients to do nothing in `CAdManager::CycleAds` without involving master server in it. It's also a nice optimization because if a server doesn't want ads, then its clients don't need to bother master server at all. But well, we can always add it if needed.

PR #281, opened 2025-02-24
Add support for tia gamerules to default maps

griseraner:


ccomrade (approved):
Hawaii will not work because it's stored in `%USER%/Downloads/Levels` and it's not a default map, but the rest is fine.

PR #286, opened 2025-04-20
Fix new ToD and enable it by default

### LINKED ISSUE #81: Indoor areas are sometimes all white with new ToD
Confirmed to happen only with the new Time of Day implementation. Reconnect and restart does not fix it.

image
image

ccomrade:
- Fix #81 by hooking one forgotten `TimeOfDay::GetHDRMultiplier` call (non-virtual inline function).
- Fix various new ToD issues in 32-bit caused by the following functions not working correctly in 32-bit:
  - `I3DEngine::GetDawnStart`
  - `I3DEngine::GetDawnEnd`
  - `I3DEngine::GetDuskStart`
  - `I3DEngine::GetDuskEnd`
- Fix subtle differences in sun shafts compared to the original ToD implementation.
- Enable new ToD by default (`-newtod` is now `-oldtod`).

Big thanks to @griseraner for nailing down the issues!

### DIFF
--- a/Code/Cry3DEngine/TimeOfDay.cpp
+++ b/Code/Cry3DEngine/TimeOfDay.cpp
@@ -486,9 +486,14 @@ void TimeOfDay::Update(bool interpolate, bool forceUpdate)
 		m_vars[NIGHT_SKY_MOON_OUTER_CORONA_SCALE].value[0]
 	);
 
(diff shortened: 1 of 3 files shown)


griseraner (approved):
Yay.. Great work!!

PR #287, opened 2025-04-21
Integrate DSOAL and OpenAL Soft

ccomrade:
## Intro

Integrate [DSOAL](https://github.com/kcat/dsoal) and [OpenAL Soft](https://github.com/kcat/openal-soft) to enable 3D surround sound without special audio hardware. DSOAL uses OpenAL Soft to emulate [DirectSound](https://en.wikipedia.org/wiki/DirectSound) (DirectX Audio), including [EAX](https://en.wikipedia.org/wiki/Environmental_Audio_Extensions).

Thanks to @ThreeDeeJay for the original idea.

## Audio path

Before:

```
CrySoundSystem -> FMOD -> DirectSound
```

Now:

```
CrySoundSystem -> FMOD -> DSOAL -> OpenAL Soft -> WASAPI
```

Modern [WASAPI](https://learn.microsoft.com/en-us/windows/win32/coreaudio/wasapi) is now used as the audio backend instead of [DirectSound](https://en.wikipedia.org/wiki/DirectSound), which is deprecated since Windows Vista.

The original `dsound.dll` (DirectSound) is not loaded at all anymore. This makes us less dependent on Windows-specific APIs, which brings us closer to making Crysis truly cross-platform. OpenAL Soft supports many platforms already.

## Configuration

It seems Crysis does not use EAX by default. Depending on sound quality settings, `s_ReverbType` cvar is either 0 (no reverb) or 2 (software reverb):

```
variable: s_ReverbType 
type: int
current: 2
help: Toggles type of reverb effect.
Usage: s_ReverbType [0..6]
Default is 2 (software).
0: Disables reverb completely.
1: Enables HW I3DL2 reverb (EAX).
2: Enables SW I3DL2 reverb.
3: Enables DSP based FreeVerb Reverb.
4: Enables DSP based VST plugin: Classic Reverb.
5: Enables DSP based Princeton2016 (disabled).
6: Enables DSP based RoomMachine844.
```

The recommended config (`system.cfg`) from @ThreeDeeJay is the following:

```
s_OutputConfig 1
s_MinHWChannels 1
s_MaxHWChannels 256
s_ReverbType 1
```

With this config, sound system logs the following during startup (`-verbosity 4`):

```
Sound initialization
CryPak::FOpen("game/Libs/ReverbPresets/ReverbPresets.xml", "rb", 0x0): 0x0000000000110022 Found in pak "game/GameData.pak"
CryPak::FClose(0x0000000000110022)
[CryMP] Using DSOAL and OpenAL for sound
 Driver supports hardware 3D sound
 Driver supports EAX 4.0 reverb
Sound - initializing AudioDevice now!
Sound - trying to initialize DirectSound output! 
Sound - starting to initialize DirectSound output! 
Sound - 3 drivers found: 
Sound - available drivers: 0 PulseAudio Output !
Sound - available drivers: 1 Ellesmere HDMI Audio [Radeon RX 470/480 / 570/580/590] !
Sound - available drivers: 2 aune usb dac Analog Stereo !
Sound - initializing FMOD-EX now!
Sound - initialized FMOD-EX
Sound - available record drivers: 0 PulseAudio Input !
Sound - available record drivers: 1 Monitor of aune usb dac Analog Stereo !
Sound - available record drivers: 2 Monitor of HDA ATI HDMI Digital Stereo (HDMI 5) !
Sound - using FMOD version: 00040723 and internal 00040723!
-------------CRYSOUND-EX VERSION =   00040723 ------- 
Total number of 2D hardware channels available: 128
Total number of 3D hardware channels available: 128
Total number of all hardware channels available: 256
Sound - initializing SoundSystem now!
```

EAX seems to be enabled and all looks good. However, this config makes in-game voices disappear completely and the following warning is constantly logged:

```
[Warning] <Sound> Sound-FmodEx-AudioDevice: event system update failed!  (57) A call to a standard soundcard driver failed, which could possibly mean a bug in the driver or resources were missing or exhausted. 
```

Tested on both Windows and Linux and with both DLL-based DSOAL and CryMP one (this PR). All combinations lead to the same results.

Moreover, we shouldn't depend on the externally provided `system.cfg` file. CryMP should configure the sound system appropriately on its own.

OpenAL Soft also has its own optional [`alsoft.ini`](https://github.com/kcat/openal-soft/blob/master/alsoftrc.sample) config file.

## Integration

DSOAL and OpenAL Soft are normally two DLLs (`dsound.dll` and `dsoal-aldrv.dll`). We cannot use them because:
- Vanilla game would be affected as well.
- CryMP is designed as a single self-contained executable for easy deployment.

This is why DSOAL and OpenAL Soft are integrated directly into the CryMP executable. DSOAL no longer loads the OpenAL Soft DLL and uses the one in the EXE instead.

Additional changes in DSOAL and OpenAL Soft source code were necessary to make them coexist in a single executable. For example, fixing collisions in global variables. Both also come with their own copy of [fmtlib](https://github.com/fmtlib/fmt) used for logging. For this reason, our DSOAL does not log anything currently.

Only source code of DSOAL and OpenAL Soft and their license files are added. Other unnecessary files from their repositories are omitted. Including their build system, which is reimplemented in our own `CMakeLists.txt` in a much simpler and cleaner way.

## TODO

- [ ] Sound system configuration issues. See above.
- [ ] It seems only some sounds take the 3D/EAX path. Reported by @ThreeDeeJay using vanilla game and DLL-based DSOAL.
- [ ] Does the EAX emulation actually work correctly in Crysis? It would be nice to make a comparison between DSOAL and hardware EAX.
- [ ] There seems to be no difference in sound with and without DSOAL under Linux (Wine + DXVK). Tested with headphones and both DLL-based DSOAL and CryMP one.
- [x] Add `-nodsoal` command line option to allow switching back the original DirectSound implementation without DSOAL and OpenAL Soft.
- [ ] Upstream at least some integration changes in DSOAL and OpenAL Soft to make updates of their code easy. Both are actively maintained. Maybe we could even add them as git submodules.
- [ ] OpenAL Soft logs to stderr instead of CryLog.
- [ ] DSOAL logs are currently disabled completely.

Other than that, everything seems to work well already.


ThreeDeeJay:
Oh dang, I guess I could've sworn I had heard voices (maybe just enemies?) despite the `call to a standard soundcard driver failed` error. I reported that and other issues mentioned like the partial 3D audio here https://github.com/kcat/dsoal/issues/124 and previously here https://github.com/kcat/openal-soft/issues/1124

>  Does the EAX emulation actually work correctly in Crysis? It would be nice to make a comparison between DSOAL and hardware EAX.

I have a Creative X-Fi USB sound card that supposedly supports hardware EAX but I suspect it's just software camouflaging as hardware, so that comparison would probably require someone to test it with an internal PCI X-Fi.
That said, I did happen to record a quick test showing EAX is being emulated by DSOAL:

https://github.com/user-attachments/assets/85438db7-e394-4ce9-a4c4-cb618073b9e2

Note: I added this to alsoft.ini to exaggerate the reverb to make it obvious in this particular clip, and you can also notice only 3D sounds get reverb.
```
[reverb]
boost=24
```

I should probably point out: I don't know if Crysis ever advertised EAX or if it was even meant to work even on Windows XP with EAX hardware (I tested it myself and just got non-3D stereo/surround). I just happened to stumble upon some magic Cry engine cvars and just got lucky. well, at least partially lol

And by the way, you might only notice the difference DSOAL makes in audio if HRTF (3D audio) is enabled. By default, it's auto (only enabled if the device is detected as headphones) though it can be forced via alsoft.ini, which is why [DSOAL](https://github.com/kcat/dsoal/releases/download/latest-master/DSOAL.zip) comes with a separate folder pre-configured for HRTF. But it should provide an advantage in both multiplayer but also single player so *if* all the issues above are figured out, maybe people would appreciate better sound positioning awareness during the campaign, even if it's disabled by default if it's ever added to the Sound settings in-game. 👀👌

mirh:
~~I'm not sure what you are talking about but that is probably related to [this](https://github.com/kcat/dsoal/issues/18#issuecomment-1467126510) and [this](https://github.com/kcat/openal-soft/issues/675#issuecomment-1075728969) to take into account that could bite our ass.
And oh boy, I'd be very surprised if FMOD actually didn't screw this up on top of the clusterfuck creative's drivers already were.~~

~~So.. eh? Maybe the best course of action would be first to try to experiment with their demos/samples lol.
As for experimenting with the dlls in crysis, the thing with my links is that the game uses 4.07.23. Admittedly the best thing would be finding out version 4.07.25 somewhere (as you can notice in the readme the other version had introduced~~

> 7/09/07 4.07.25 - Dev release update
>\-------------------------------------
> \- Fix 2d/3d pan morphing from not working properly with stereo and above sounds.

Lmao, from the same readme.
So, anyway, as I was saying.. Ideally you'd want to try with the immediately before or after version, or if not any at least the same major version, but those couldn't be archived (the demo ships the same version btw).
Hence it would be interesting to see if 4.06.12 and 4.08.06 are near enough to still work, or if they are going to blow up the linking.

ThreeDeeJay:
@mirh Sorry for the late response, but I just extracted `api/fmodex.dll` (and `fmoddesignerapi/api/fmod_event_net.dll and fmod_event_net.dll` or else I'd get no sound) from [4.08.06](https://web.archive.org/web/20071027022757/http://www.fmod.org/index.php/release/version/fmodapi40806win64-installer.exe) into `Crysis/Bin64` but it didn't fix certain sounds not being treated as 3D/"hardware". 
```
Sound initialization
File open: game/Libs/ReverbPresets/ReverbPresets.xml
 Driver supports hardware 3D sound
 Driver supports EAX 4.0 reverb
Sound - initializing AudioDevice now!
Sound - trying to initialize DirectSound output! 
Sound - starting to initialize DirectSound output! 
Sound - 1 drivers found: 
Sound - available drivers: 0 Speaker (SB X-Fi Surround 5.1) !
Sound - initializing FMOD-EX now!
Sound - initialized FMOD-EX
Sound - using FMOD version: 00040806 and internal 00040723!
-------------CRYSOUND-EX VERSION =   00040723 ------- 
Total number of 2D hardware channels available: 128
Total number of 3D hardware channels available: 128
Total number of all hardware channels available: 256
Sound - initializing SoundSystem now!
File open: game/Libs/SoundMoods/SoundMoods.xml
File open: game/Libs/ReverbPresets/ReverbPresets.xml
```
[Crysis+FMOD_4.08.06_logs.zip](https://github.com/user-attachments/files/24353122/Crysis%2BFMOD_4.08.06_logs.zip)

And even though I didn't get this error anymore:
```
[Warning] <Sound> Sound-FmodEx-AudioDevice: event system update failed!  (57) A call to a standard soundcard driver failed, which could possibly mean a bug in the driver or resources were missing or exhausted. 
```
now I'm getting a bunch files failing to load like this:
```
File open: languages/dialog/island/jester_island_ab3_00080.mp2
[Warning] <Sound> SB: create sample sound failed! languages/dialog/island/jester_island_ab3_00080.mp2 (37) An invalid parameter was passed to this function. 
<Sound> Error: Sound - Cannot load sample sound languages/dialog/island/jester_island_ab3_00080.mp2
[Warning] <Sound> Sound languages/dialog/island/jester_island_ab3_00080.mp2 failed to load!
```

I also tried the same with the [32-bit DLLs](https://web.archive.org/web/20160305023639*/http://www.fmod.org/index.php/release/version/fmodapi40806win32-installer.exe) but got errors and no sound.
csrss_DyldScH6Vd
Crysis_BKxMXPfGCo

- I also tried [40612win64](https://web.archive.org/web/20070317223723/http://www.fmod.org:80/files/fmodapi40612win64-installer.exe) but I get no sound, maybe because `fmod_event_net.dll` was missing. And if I just use the original one, I get these errors and still no sound:
csrss_fNmYE62RHZcsrss_tAzv7mtkJE

-  [40612win32](https://web.archive.org/web/20070312005557/http://www.fmod.org:80/files/fmodapi40612win32-installer.exe) did include `fmod_event_net.dll` but it just hung with a black screen and the game cursor.
csrss_jkUiAgde89

mirh:
I assume 64-bit 4.12.09 is far too far fetched to try..

> 18/10/07 4.08.06 - Stable release update
> \----------------------------------------
> \+ Added support for Channel::setPan and Channel::setSpeakerLevels on 3d voices
  so that they work when Channel::set3DPanLevel is used, as it should according
  to the documentation.  (only Channel::setSpeakerMix was working).

Also found this in the respective readme. I don't know why it's not mentioned anywhere else.

ThreeDeeJay:
I'm getting not found errors on these pages.
https://web.archive.org/web/20210403215822/https://www.fmod.com/index.php/release/version/fmodapi41209win32-installer.exe
https://web.archive.org/web/20210403215820/https://www.fmod.com/index.php/release/version/fmodapi41209win64-installer.exe
Same for
https://web.archive.org/web/20140801013046/http://www.fmod.org/release/version/fmodapi41403win32-installer.exe
https://web.archive.org/web/20140801013046/http://www.fmod.org/release/version/fmodapi41403win64-installer.exe
https://web.archive.org/web/20210405020234/https://www.fmod.com/index.php/release/version/fmodapi41810win32-installer.exe
https://web.archive.org/web/20210402200224/https://www.fmod.com/index.php/release/version/fmodapi42204win32-installer.exe
https://web.archive.org/web/20210403113209/https://www.fmod.com/index.php/release/version/fmodapi42409win32-installer.exe
https://web.archive.org/web/20240524162752/https://www.fmod.com/index.php/release/version/fmodapi42636win32-installer.exe
https://web.archive.org/web/20210413082132/https://www.fmod.com/index.php/release/version/fmodapi42809win32-installer.exe

mirh:
FMOD Designer was to be installed on top of the Ex API, so.. yeah.

> 6/12/07 4.10.00 - Initial stable release 
> \-----------------------------------------
> \* .FEV file format updated! FMOD Designer 4.11.00 requires FMOD version 4.11.00,
  4.10.00, 4.08.10 or higher.

And I wonder if this couldn't be the reason that you stopped to get sounds there (even 4.08.00 had such a breaking change tbh, but maybe they still used to retain backward compatibility for at least a single older version).

So, anyway.. idk. If I was you I'd try to check with some [older](https://www.mobygames.com/group/13947/sound-engine-fmod/include_dlc:true/sort:date/) FMOD 4 [game](https://web.archive.org/web/20071106080253/http://www.fmod.org/), if in that year or so between Ex becoming production ready and FMOD_HARDWARE dropping mixing at least _some_ is how you'd expect it.
If so then we may have a chance tinkering with newer dlls to have the 3D space be exposed externally (like, I dunno.. maybe it's just about overriding `DSBCAPS_LOCSOFTWARE`?) otherwise if it never worked to begin with it's all a fruitless endeavour.
Adding openAL to the sources of the remaster is probably a better idea, or maybe you could do that with crysis 2's CrySoundSystem and then try to backport the dll to this.

diznq:
I think that if voices are the only thing missing, this could perhaps be merged? After all, this is a multiplayer client, so a) there are no enemy voices to be heard to begin with, and b) if this affects in-game radio voices, it's only good, those voices are annoying anyway 😄

diznq (approved):

ccomrade:
I agree that we should understand (and fix) those remaining issues. Implementing our own CrySoundSystem is the way. It shouldn't be that much effort because CrySoundSystem seems to be mostly just a wrapper around FMOD. And we can choose the FMOD version! The modern FMOD looks nice. They even offer a non-commercial license for free. But is this modern FMOD still compatible with Crysis sound assets? Does anyone have any experience with this?

PR #290, opened 2025-04-27
Weapon attachment menu improvements

griseraner:
- Faster weapon customization possible
- Added animation blending 
- Enabled animation as seen in the early Alpha HUD
- Fixed some bugs in the flash file


ccomrade (approved):
Nice :+1:

PR #292, opened 2025-05-19
Fix attachment menu bugs

griseraner:


ccomrade (approved):
Looks and works good.

PR #293, opened 2025-05-25
Completly remove aim assistant implementation from game

diznq:
There is an ongoing debate of accusations of several highly skilled players using aim assistant to cheat their way through.
Yes, indeed, it is possible even if server disables it:

- they can plug joy-stick into computer to make the following condition true even if they use keyboard and mouse `gEnv->pInput->HasInputDeviceOfType(eIDT_Gamepad) && gEnv->pTimer->GetCurrTime() - m_lastNonAssistedInput > g_pGameCVars->aim_assistRestrictionTimeout`
- they can use CheatEngine or similar tools to force set all `aim_*` values to predefined configuration (yeah, it works in MP, I tested it and I was indeed an aim god... pls don't judge, it was for scientific research 😂😂)
- they can abuse `buy ...` exploit in console to execute custom Lua code and call `CPPAPI.FSetCvar`

By removing this code, we can finally settle down the debate. Can't use aim assistant if there is no assistant to begin with 😎 And as for players who refuse to update their game... well that will be quite telling as to why 😬


ccomrade (approved):
The drama continues... :smile:

I'm pretty sure nobody has ever pulled this off. As already mentioned, we can check client cvars remotely. Server is even informed about active aim assist via `SvRequestHitAssistance` RMI.

There's this toxic guy that insults and accuses others of using aim assist. He claims that it's possible to activate aim assist in multiplayer by putting `aim_*` cvars into `system.cfg`, which is nonsense.

However, I agree that it's better to remove aim assist completely. Crytek had a good intention in making the game more playable for gamepad players, but it doesn't make much sense anyway. Especially in multiplayer.

The changes look good. Thanks.

PR #294, opened 2025-05-25
Replace VeraMono with DejaVuSansMono

ccomrade:
Our CryFont already supports UTF-8 (#138). However, the original VeraMono font is very limited.

This PR replaces the original VeraMono font with DejaVuSansMono. It allows CryFont to render Unicode characters and symbols.

Localized ranks in name tags are now rendered correctly:

### Before

image

### After

image

Also, the `INetwork::GetService` log message is now hidden by default. This is not related to the font change.


griseraner:
Looks nice

PR #296, opened 2025-06-08
Fix watchtower

griseraner:


ccomrade (approved):
It's possible to hit players inside the watchtower now. Nice!

ScreenShot0064

PR #300, opened 2025-06-29
fix: Module not found issue

diznq:
Replace `std::chrono::clock_cast` because it throws `std::runtime_error` on old Windows.


ccomrade:
Thanks!

PR #313, opened 2025-07-20
Use temporary unique filenames during download

ccomrade:
## The problem

When a file download is canceled by the user and then quickly retried, it may fail. This is because the cancellation of the previous request has not completed yet, so the destination file is still open and the new request is not able to reopen it.

Thanks to Czarny_Msciciel for reporting this issue.

## The fix

To fix this issue, a random number is added into the destination filename. Each download request now has its own unique filename. Additionally, the destination file is now opened after the HTTP connection is established, not before.

## Other changes

- Replace `WinAPI::File` with `StdFile`, which is a simple `std::FILE` wrapper.
- Remove unnecessary `ServerPAK::IsZipFile` as we no longer use the original CryPak.

## Testing

```
[CryMP] Checking server at 185.22.237.253:11522
[CryMP] [MapDownloader] Downloading map multiplayer/ps/recovery
[CryMP] [MapDownloader] Version 2025.3
[CryMP] [MapDownloader] From http://crymp.org/static/Recovery202503.zip
[CryMP] [MapDownloader] To Z:\home\comrade\Documents\My Games\Crysis\Downloads\multiplayer_ps_recovery_563157489.zip
[CryMP] [MapDownloader] Download finished, extracting...
[CryMP] [MapDownloader] Map extracted, checking map folder...
[CryMP] [MapDownloader] Map downloaded successfully
[CryMP] Obtaining server PAK...
[CryMP] [FileCache] Downloading from https://crymp.org/static/ZVibes202502.pak
[CryMP] [FileCache] Download finished
[CryMP] [ServerPAK] Loaded Z:\home\comrade\Documents\My Games\Crysis\Downloads\Cache\54F8BADE57F45A609C70EA93FA31999F01426BBD65AC72CE8DB3521A39294F99
[CryMP] Joining VIBING 24/7 at 192.168.0.2:11522
```


griseraner (approved):
Looks good, nice fix

PR #316, opened 2025-07-27
fix: Use XML weapon and ammo definitions when in server mode

diznq:


griseraner:
👍 Maybe the same for vehicle xmls?

Code/CryGame/Items/Weapons/WeaponSystem.cpp
```diff
@@ -456,6 +463,142 @@ int  CWeaponSystem::QueryProjectiles(SProjectileQuery& q)
...
+			{
+				CryLogWarningAlways("Invalid XML file '%s'! Skipping...", xmlFile.c_str());
+				continue;
+			}
+
+			if (!ScanXML(rootNode, xmlFile.c_str()))
+				continue;
+
+		} while (pPak->FindNext(handle, &fd) >= 0);
+	}
```
ccomrade, line 518: The original code from Mod SDK leaks CryPak search handles due to missing `ICryPak::FindClose` call. Better to use our own fancy `CryFind`:

```suggestion
	if (!m_recursing)
		CryLog("Loading ammo XML definitions from '%s'!", folderName);

	for (auto& entry : CryFind(search.c_str()))
	{
		if (entry.IsDirectory())
		{
			string subName = folder + "/" + entry.name;
			if (m_recursing)
			{
				Scan(subName.c_str());
			}
			else
			{
				m_recursing=true;
				Scan(subName.c_str());
				m_recursing=false;
			}
			continue;
		}

		if (_stricmp(CryPath::GetExt(entry.name), "xml"))
		{
			continue;
		}

		string xmlFile = folder + string("/") + string(entry.name);
		XmlNodeRef rootNode = m_pSystem->LoadXmlFile(xmlFile.c_str());

		if (!rootNode)
		{
			CryLogWarning("Invalid XML file '%s'! Skipping...", xmlFile.c_str());
			continue;
		}

		ScanXML(rootNode, xmlFile.c_str());
	}
```

ccomrade (changes requested):
Hmmm the original messy weapon XML loader code is back. It's fine for now, until we come up with a better solution instead of editing XML files. I would just fix that search handle leak.

PR #318, opened 2025-08-02
Obstruct object crash fix

griseraner:


ccomrade (approved):
Looks good. Nice catch!

PR #321, opened 2025-08-04
Temporary fix for invisible objects

griseraner:


ccomrade (approved):
Great job finding this! Although it's only a workaround. We still need to figure out what the actual issue is and fix it properly. But yes, this is fine as a temporary solution.

PR #323, opened 2025-08-08
Add new US nanosuit helmet mask

ccomrade:
image

## Before

image

## Now

image

Thanks to Sagitta.


Sagitta11:
My proposal would be to put the vanilla helmet or my fix helmet as an addon.(Just like Half life 1 hd content)

ccomrade:
> My proposal would be to make the vanilla helmet or my fix helmet as an addon.(Just like Half life 1 hd content option) If the creators of the crymp client approve this, i recommend the vanilla helmet should be an addon.(I recommend that because i say once more that the second option won)

Shouldn't it be the other way around in that case? Vanilla being the default and if someone wants, they can use the new one?

Sagitta11:
That works too.

Sagitta11:
So what do we do now?

ccomrade:
True, client pak has the highest priority. Not even server pak can override it. We should change that probably. And yes, replacing textures on the fly without game restart is tricky right now. I agree that we need to implement reliable handling of textures in server pak first.

## Conclusion

Currently, servers and players have no way to restore the original helmet if we merge this. If we keep the default, it's still possible to use the new one if someone wants. Closing.

PR #330, opened 2025-08-22
feat: Add UTF-8 Chat

diznq:
Add support for writing into chat in various alphabets, such as cyrillic or extended latin.


griseraner:
Nice, but I thought we need to edit flash files also? 😃

Code/Library/WinAPI.cpp
```diff
@@ -610,6 +611,108 @@ wchar_t WinAPI::WideCharToUpper(wchar_t ch, int languageID)
 	}
 }
 
+std::wstring WinAPI::CharToWString(char c) {
+	WCHAR wideChar[2];
+	int wlen = MultiByteToWideChar(GetActiveKeyboardAnsiCP(), 0, &c, 1, wideChar, 2);
```
ccomrade, line 616: Is it a good idea to query current keyboard layout and its code page for *every* input character?

By the way, believe it or not, input key events are dispatched in our code:

https://github.com/crymp-net/crymp-client/blob/83019074e2dfb8ee8290ad555a225cc807660ed6/Code/CrySystem/GameWindow.cpp#L56-L89

It's the reverse engineered original implementation. We can fix it and avoid this conversion back to wide char completely. The wide char is actually passed through `event.timestamp`. And `WM_INPUTLANGCHANGE` can be used to effectively detect keyboard layout changes without querying it.

What do you think?

Code/CryGame/HUD/HUDTextChat.cpp
```diff
@@ -374,16 +378,16 @@ void CHUDTextChat::Down()
...
-void CHUDTextChat::Insert(char ch)
+void CHUDTextChat::Insert(wchar_t ch)
 {
 	if (m_inputText.length() >= MAX_MESSAGE_LENGTH)
 	{
 		return;
 	}
 
-	if (!std::isprint(ch))
+	if (ch == 0)
```
ccomrade, line 388:
```suggestion
	if (!std::iswprint(ch))
```

ccomrade (approved):
Looks good. Thanks!

PR #339, opened 2025-08-30
Replace GameSpy logo in lobby

ccomrade:
image


ccomrade:
Credit for the new logo goes to Payro.

PR #343, opened 2025-09-06
Add LocalizationManagerInfo console command and fix missing @ui_E

ccomrade:
Add `LocalizationManagerInfo` console command for debugging localization issues.

Fix map coordinate "E" incorrectly shown as "B" in Russian localization:

image

The issue was that we didn't add `@ui_E` label because it already exists. However, the original label is meant as **E**ast, which is **В**осток (?) in Russian. ~~I'm not aware of any place where the original label is used.~~ Therefore, the fix is to set it to "E" in all languages. Both english and localized text needs to be set in this case because the original label sets localized text, which has higher priority when set.

Thanks to MAIKI_FG for reporting this bug!


ccomrade:
Thanks for the fix!

PR #344, opened 2025-09-15
72 sometimes rockets go through staticterrain

griseraner:
* Fix MP projectile explosion FX not playing because the server removed projectiles before the client registered a collision. Effects/decals now trigger from the **ClExplosion** event instead of per-particle client collisions.

**Changes**

* Added `CGameRules::PlayMFXFromExplosionInfo` (plus small helpers) to drive MaterialFX from `ClExplosion` with a short forward probe for surface/material.
* Imported `OnCollisionLogged_MaterialFX` from CryAction, and try to pass correct **pt/normal/velocities/materials**.
* Actor hits: ensure the normal faces against shot direction for living targets so backface test doesn’t cull the effect.
* Water-aware resolve (prefer water plane if it’s in front of the solid) and cleaner decal projection (no more smeared/oversized ground decals).
* Suppression of legacy client per-particle collision FX when using ClExplosion mfx (avoids duplicates).
* New cvar **`mp_explosion_mfx`** (on/off).
* Ammo flag **`clexplosion_mfx`** (XML) for projectiles that should use the ClExplosion MFX path.

**Behavior**

* In MP (and SP), explosion FX/decals now play reliably even when the projectile entity is removed early by the server.
* If `mp_explosion_mfx=0` or an ammo lacks `clexplosion_mfx`, falls back to legacy behavior.

### DIFF
--- a/Code/CryGame/GameRulesClientServer.cpp
+++ b/Code/CryGame/GameRulesClientServer.cpp
@@ -27,6 +27,7 @@ History:
 #include "CryCommon/CryAction/IMaterialEffects.h"
 #include "CryCommon/CryAction/IGameplayRecorder.h"
 #include "CryCommon/CryEntitySystem/EntityId.h"
+#include "CryCommon/CrySystem/IConsole.h"
 
 #include "Items/Weapons/Weapon.h"
 #include "Items/Weapons/WeaponSystem.h"
@@ -39,6 +40,8 @@ History:
 #include "Library/Util.h"
 
 #include "CryMP/Server/SSM.h"
+#include "Items/Weapons/Projectiles/Bullet.h"
+#include "CryCommon/Cry3DEngine/IFoliage.h"
 
 
 //------------------------------------------------------------------------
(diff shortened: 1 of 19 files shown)


Code/CryGame/Items/Weapons/Projectile.h
```diff
@@ -26,7 +26,7 @@ Description: Projectile
 #include "TracerManager.h"
 #include "Weapon.h"
 #include "AmmoParams.h"
-
+#include "CryGame/GameCVars.h"
```
ccomrade, line 29: It would be better to move `IsPlayingMfxFromClExplosion` to the .cpp file so we don't have to include this here.

Code/CryGame/Items/Weapons/WeaponSystem.cpp
```diff
@@ -866,13 +866,15 @@ void CWeaponSystem::DumpGhostProjectiles()
 		// resolve owner name
 		const IEntity* pOwnerEnt = gEnv->pEntitySystem->GetEntity(p->GetOwnerId());
 		const char* ownerName = pOwnerEnt ? pOwnerEnt->GetName() : "<no owner>";
+		IGameObject* pGameObject = p->GetGameObject();
+		const uint8_t slotEnables = pGameObject ? pGameObject->GetUpdateSlotEnables(p, eIUS_General) : 0;
```
ccomrade, line 870:
```suggestion
		const int slotEnables = pGameObject ? pGameObject->GetUpdateSlotEnables(p, eIUS_General) : 0;
```

Code/CryGame/GameRulesClientServer.cpp
```diff
@@ -761,8 +799,680 @@ void CGameRules::ProcessExplosionMaterialFX(const ExplosionInfo& explosionInfo)
...
+	// Source binding
+	c.pEntity[0] = peSrc;
+	c.iForeignData[0] = pSrcEnt ? PHYS_FOREIGN_ID_ENTITY : 0;
+	c.pForeignData[0] = pSrcEnt;
+
+	c.penetration = 0.f;
+	c.normImpulse = 0.f;
+	c.radius = 0.f;
+
+	OnCollisionLogged_MaterialFX((const EventPhys*)&c);
```
ccomrade, line 1029: Would it be beneficial to call the original function in CryAction instead? It can also be done.

ccomrade (approved):
Very cool!

Code/CryGame/GameRulesClientServer.cpp
griseraner: We could try that, yes

PR #345, opened 2025-09-15
Remove SUIWideString + Entity name instead of "(null)"

ccomrade:
- Fix yet another broken UTF-8 -> UTF-16 conversion in Flash code.
- Show entity name instead of "(null)" in obituary if possible.


griseraner:
Nice fix!

PR #352, opened 2025-11-13
feat: Add server pak for servers

diznq:
Issue:

Although client supports server pak already, servers have no way to override XML files and so on

Fix:

Introduce `-pak` flag for servers

### DIFF
--- a/Code/CryMP/Server/Server.cpp
+++ b/Code/CryMP/Server/Server.cpp
@@ -15,7 +15,6 @@
 #include "Server.h"
 
 #include "CryMP/Common/ScriptBind_CPPAPI.h"
-#include "CryMP/Common/ServerPAK.h"
 
 #include "CryMP/Server/SSM.h"
 #include "CryMP/Server/SafeWriting/SafeWriting.h"


Code/CryMP/Server/Server.cpp
```diff
@@ -31,9 +32,17 @@ void Server::Init(IGameFramework* pGameFramework)
...
 	this->pHttpClient = std::make_unique<HTTPClient>(*this->pExecutor);
+	this->m_pServerPAK = std::make_unique<ServerPAK>();
 
 	pGameFramework->RegisterListener(this, "crymp-server", FRAMEWORKLISTENERPRIORITY_DEFAULT);
 
+
+	const std::string serverPak(WinAPI::CmdLine::GetArgValue("-pak"));
+	if (!serverPak.empty()) {
+		CryLogAlways("$6[CryMP] Loading server pak: %s", serverPak.c_str());
+		m_pServerPAK->Load(serverPak.c_str());
```
ccomrade: I think we can call just `CryPak::LoadServerPak` here. `ServerPAK` class should stay in the client because it has a different purpose (loading and unloading paks on the fly).

```suggestion
        const bool success = CryPak::GetInstance().LoadServerPak(serverPak);
        CryLogAlways("$6[CryMP] Loading server pak '%s' %s", serverPak.c_str(), success ? "succeeded" : "failed");
```

PR #360, opened 2025-11-30
Allow AI.RegisterWithAI in multiplayer

ccomrade:
Even though the restriction preventing CryAISystem from working in multiplayer has already been removed, the `AI.RegisterWithAI` Lua function still didn't work. It's fixed now. The `AI` table functions are actually implemented in CryAction instead of CryAISystem.

Thanks to [akeeperctl](https://github.com/akeeperctl) for pointing out this issue!


griseraner (approved):
nice

PR #362, opened 2025-12-05
Fix most of the compiler warnings + Add Lua crash callstack

diznq:
- Fix most of the warnings raised by VS static code analyzer (except uninitialized variables in CE2 header files)

### DIFF
--- a/Code/CryScriptSystem/ScriptSystem.cpp
+++ b/Code/CryScriptSystem/ScriptSystem.cpp
@@ -811,7 +811,6 @@ bool ScriptSystem::GetGlobalAny(const char *key, ScriptAnyValue & any)
 	}
 
 	memcpy(key_buffer, key, key_length + 1);
-	key_buffer[255] = 0;
 	char* current_key = key_buffer;
 
 	bool success = false;


Code/Cry3DEngine/TimeOfDay.cpp
ccomrade: Please revert all changes in this file. There's nothing wrong with ignoring `sscanf` return value when all output variables are properly initialized. And we better avoid potential bugs caused by subtle changes in ToD spline deserialization (the `break` statement).

Code/CryCommon/CryAISystem/IAgent.h
```diff
@@ -1033,13 +1033,13 @@ struct IAIObject:
...
-			int _fastcast_CPipeUser : 1;
-			int _fastcast_CPuppet : 1;
-			int _fastcast_CAIVehicle : 1;
+			unsigned int _fastcast_CAIActor : 1;
+			unsigned int _fastcast_CAIAttribute : 1;
+			unsigned int _fastcast_CAIPlayer : 1;
+			unsigned int _fastcast_CLeader : 1;
+			unsigned int _fastcast_CPipeUser : 1;
+			unsigned int _fastcast_CPuppet : 1;
+			unsigned int _fastcast_CAIVehicle : 1;
```
ccomrade: Are you sure signed and unsigned bit fields are exactly the same under MSVC?

Code/CryMP/Client/ServerBrowser.cpp
ccomrade: Please revert changes in this file as well. Especially using "127.0.0.1" as an invalid IP address.

Code/CryScriptSystem/ScriptSystem.cpp
```diff
@@ -22,6 +22,8 @@ extern "C"
 #include "ScriptTable.h"
 #include "ScriptUtil.h"
 
+static lua_State *g_L = nullptr;
```
ccomrade: Can we use `gEnv->pScriptSystem` instead of adding globals into script system again? Also, this should be in a separate PR.

Code/CryScriptSystem/ScriptSystem.cpp
```diff
@@ -756,6 +758,7 @@ void ScriptSystem::SetGlobalAny(const char *key, const ScriptAnyValue & any)
 	}
 
 	memcpy(key_buffer, key, key_length + 1);
+	key_buffer[255] = 0;
```
ccomrade: It already is null-terminated since we copy `key_length + 1`.

```suggestion
```

Code/CrySystem/CryPak.cpp
```diff
@@ -382,25 +382,29 @@ ICryPak::PakInfo* CryPak::GetPakInfo()
 	const unsigned int pakCount = m_paks.GetActiveCount();
 
 	PakInfo* info = static_cast<PakInfo*>(std::calloc(1, sizeof(PakInfo) + (sizeof(PakInfo::Pak) * pakCount)));
-	info->numOpenPaks = pakCount;
+	if (info) {
```
ccomrade: The following is way more readable:

```cpp
if (!info)
{
    return nullptr;
}
```

Code/Library/CrashLogger.cpp
```diff
@@ -23,6 +23,8 @@ static CrashLogger::LogFileProvider g_logFileProvider;
 static CrashLogger::HeapInfoProvider g_heapInfoProvider;
 static const char* g_banner;
 
+void DumpLuaStackTrace(std::FILE *file);
```
ccomrade: Library now depends on CryScriptSystem...

Pak/CryMP/Scripts/Entities/vehicles/VehicleBase.lua
ccomrade: Please move this into a separate PR because I really don't like when a PR for fixing compiler warnings changes other stuff.

Code/CryCommon/CryAISystem/IAgent.h
diznq: Yes, all the other CryEngine code uses unsigned bits anyway

Code/Library/WinAPI.cpp
```diff
@@ -785,6 +785,10 @@ int WinAPI::HTTPRequest(
 		}
 	}
 
+	if (!urlComponents.lpszHostName) {
+		throw StringTools::SysErrorFormat("Invalid Hostname");
+	}
+
```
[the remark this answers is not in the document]
diznq: I don't know, but MSVC thinks it can happen 😄

PR #380, opened 2026-01-04
Pause sound when unfocused window

griseraner:


ccomrade (approved):
Looks good. It even compiles without adding:

```cpp
#include "CryCommon/CrySoundSystem/ISound.h"
```

Included by some other header it seems.

PR #384, opened 2026-01-18
Restrict problematic cvars in multiplayer

ccomrade:
There are cvars that are not net-synced and can break the game. This is now fixed in multiplayer.

#### Before

ScreenShot0142

#### Now

ScreenShot0143


griseraner (commented):
Looks good, although maybe the other cvars can use SetOnChangeCallback as well, maybe make a Client::ProtectCVar or something, then we could add CPPAPI function so servers can add additional cvars if needed

PR #387, opened 2026-02-08
Add client pak hot reloading option and CMake Presets

ccomrade:
This PR introduces the `CRYMP_PAK_HOT_RELOADING_ENABLED` build option. When enabled, the `Pak/` folder content within the source code is used directly. This allows, for example, hot reloading of internal Lua scripts for rapid development. No need to rebuild and restart the client each time. Live editing is now possible.

A small optimization of `CryPak::SearchWildcardPath` duplicate detection is also included.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -46,9 +46,9 @@
 			}
 		},
 		{
-			"name": "hot-reloading-bin32",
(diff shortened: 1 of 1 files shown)


griseraner:
Great!

griseraner (approved):

PR #389, opened 2026-02-13
feat: Improve chat system

diznq:
Add support for the following behaviors:

- chat target rotation using TAB (rotates in this order: ALL -> TEAM -> Selected players)
- PM system directly integrated within client
- reuse last chat target when in PM mode
- hide chat from muted players
- hide chat completly from other players (`cl_hud_chat 0` - yes, it is `cl_` because people are already familiar  with `cl_hud`)

As for the new chat system, it's behavior can be controlled using `mp_chat` which defaults to `1`.

These are the options:

`mp_chat 0`: Completly disables new chat behavior, using old chat
`mp_chat 1 (default)`: Enables new chat, but uses `!pm TargetName Message` for sending messages (server doesn't specify support in this case, so !pm is the safest way to not break old servers)
`mp_chat 2 (recommended)`: Enables new chat with server partially supporting it (by default servers don't send outgoing message back to originator when target is eChatTarget, so client forces displaying it)
`mp_chat 3`: Enables new chat with full server support (client expects server will handle sending the chat message back to originator)


ccomrade:
Screenshot From 2026-02-14 12-47-40

PR #417, opened 2026-04-04
fix: Player list for custom game modes

diznq:


ccomrade (changes requested):

Code/CryGame/Menus/MultiplayerMenu.cpp
```diff
@@ -56,10 +60,11 @@ enum EServerInfoKey
 
 	//crymp
 	eSIK_connectable,
+	eSIK_teams,
```
ccomrade: `eSIK_teams` is unused. Now i realized that we can actually use it to pass the metadata in a simple and clean way. Just like [`eSIK_connectable`](https://github.com/crymp-net/crymp-client/blob/68f057e8d6655759767cd6cbec62a547b8ab7379/Code/CryMP/Client/ServerBrowser.cpp#L152).

Code/CryGame/Menus/MultiplayerMenu.cpp
```diff
@@ -27,6 +27,10 @@ Description: Multiplayer lobby
 #include "CryMP/Client/Client.h"
 #include "CryMP/Client/ServerBrowser.h"
 
+#include <nlohmann/json.hpp>
+
+using json = nlohmann::json;
+
```
ccomrade:
```suggestion
```

ccomrade (approved):
Looks good

PR #421, opened 2026-04-29
Fixed invisible FP weapon after parachute

griseraner:
### Fixes
- Fixed first-person weapons becoming invisible after parachute landing
  - Regression introduced by FP/TP animation persistence changes
  - Parachute forced a deselect using a frozen final animation frame
  - Previous weapon could remain stuck in an offscreen pose when reselected
  - Now clears FP animation state during parachute switch


ccomrade (approved):
Nice

PR #422, opened 2026-04-30
Fix DOF blur not working in modify mode; use CPlayer DOF system

griseraner:
Closes #407

Fix DOF blur not working in modify mode; use CPlayer DOF system

- Fix DOF blur being overridden and not applied
- Use CPlayer DOF system (ported from Lua) instead of separate weapon logic
- Restore smooth blur fade when entering/exiting modify mode
- Fix missing DOF blur when ADS with iron sights and reflex sights
- Minor cleanup and removal of redundant DOF handling
- Replace scheduler helper classes with lambdas


ccomrade (approved):
Looks good!

PR #427, opened 2026-05-12
Replace remote held entity mass workaround with physics flag

griseraner:
This update replaces the old mass `0` workaround used for remote held entities in multiplayer with a proper CryPhysics-side fix.

Previously, remote held objects had their mass temporarily set to `0` to stop incoming physics snapshots from fighting the attachment system. While this mostly worked, it could occasionally cause desync issues, floating objects, and vehicle physics problems that sometimes required forced rephysicalization after release.

This change adds a dedicated `pef_ignore_network_state` physics flag and integrates it into `CRigidEntity::SetStateFromSnapshot()`.

When the flag is enabled, incoming rigid-body network state is ignored while the object is attachment-driven by a remote player.

## Changes

* Added new `pef_ignore_network_state` physics flag
* Prevents incoming network physics snapshots from overriding attachment-driven transforms
* Fixed local ignore-buddy constraints being removed by network state restoration
* Replaced the old mass `0` workaround in `COffHand`
* Removed unnecessary vehicle rephysicalization workaround

## Result

Remote held entities and vehicles now behave correctly in multiplayer without needing temporary zero-mass physicalization.


ccomrade (approved):
Amazing!

PR #428, opened 2026-05-13
Fix crashes and hangs caused by oversized breakable grids

griseraner:
- Fixed crashes and hangs caused by extremely large breakable grids on some custom maps
- Replaced recursive flood fill with an iterative version to avoid stack overflows
- Added a configurable limit for breakable grid size in PhysicsVars
- Disabled chunk generation for oversized breakable grids
- Improved stability when breaking problematic map objects

### DIFF
--- a/Code/CryPhysics/BreakableGrid2d.cpp
+++ b/Code/CryPhysics/BreakableGrid2d.cpp
@@ -1,4 +1,5 @@
 #include "BreakableGrid2d.h"
+#include "CryCommon/CrySystem/ISystem.h"
 #include "Utils.h"
 
 int CBreakableGrid2d::get_neighb(int iTri, int iEdge)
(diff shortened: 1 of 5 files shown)


Code/CryPhysics/BreakableGrid2d.cpp
```diff
@@ -182,8 +209,13 @@ void CBreakableGrid2d::Generate(vector2df* ptsrc, int npt, const vector2di& nCel
 int* CBreakableGrid2d::BreakIntoChunks(const vector2df& pt, float r, vector2df*& ptout, int maxPatchTris,
                                        float jointhresh, int seed)
 {
+	ptout = 0;
```
ccomrade:
```suggestion
	ptout = nullptr;
```

Code/CryPhysics/BreakableGrid2d.cpp
```diff
@@ -182,8 +209,13 @@ void CBreakableGrid2d::Generate(vector2df* ptsrc, int npt, const vector2di& nCel
...
+
+	if (m_disabled)
+		return 0;
```
ccomrade:
```suggestion
		return nullptr;
```

ccomrade (approved):
Looks good!

PR #429, opened 2026-05-15
Fix articulated ragdolls freezing mid-air

griseraner:
## Changes

- Fixed player ragdolls occasionally freezing mid-air after death
- Moved the fix into the proper CryPhysics sim class place inside `RepositionEntity()`
- Prevent airborne ragdolls from transitioning from active (`2`) to sleeping (`1`) sim class
- Removed the previous workaround that force-kept ragdolls awake every frame in `CActor::Update`
- Grounded ragdolls still sleep normally and can be woken up by impacts/interactions


ccomrade (approved):
Nice!

PR #431, opened 2026-05-31
Fix crash in rigid projectile surface mapping

griseraner:
Fixed a crash caused by invalid material mapping access on rigid-body
projectiles. Added proper CryPhysics structure initialization and
validation before accessing pe_params_part data, preventing crashes
for explosives such as Claymores, C4, and AV Mines.

### DIFF
--- a/Code/CryGame/Items/Weapons/Projectile.cpp
+++ b/Code/CryGame/Items/Weapons/Projectile.cpp
@@ -104,31 +104,44 @@ bool CProjectile::SetAspectProfile(EEntityAspects aspect, uint8 profile)
 		case ePT_Rigid:
 		{
 			SEntityPhysicalizeParams params;
+			memset(&params, 0, sizeof(params));
+
 			params.type = PE_RIGID;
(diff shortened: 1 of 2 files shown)


Code/CryGame/Items/Weapons/Projectile.cpp
```diff
@@ -104,31 +104,44 @@ bool CProjectile::SetAspectProfile(EEntityAspects aspect, uint8 profile)
 		case ePT_Rigid:
 		{
 			SEntityPhysicalizeParams params;
+			memset(&params, 0, sizeof(params));
+
```
ccomrade: Better to use `{}` instead of `memset`:
```suggestion
			SEntityPhysicalizeParams params{};
```
The same applies to other classes as well.

ccomrade (approved):

Code/CryGame/Items/Weapons/Projectile.cpp
ccomrade: 611d62e0b1bced785406b2281e4f46328ec2be20

PR #432, opened 2026-06-02
Improved cloak shader

SagittaAsus:
Cloak1
Cloak3
Cloak2

I changed the intensity of the light. It looks different when you play with it.


diznq:
I don't think this belongs to the client. Shaders should be vanilla in client and it should be up to server to provide a different resource in side server pak.

SagittaAsus:
> I don't think this belongs to the client. Shaders should be vanilla in client and it should be up to server to provide a different resource in side server pak.

 Original cloak is completly useless in multiplayer. Nobody uses cloak mode (besides of the server who has an improved version). It's very easy is to spot someone.  The cloak shader is more visible when the server uses TOD and It makes it more unusable. I think we need to improve the vanilla shader. At least for now i want to replace and see if the players prefer the new one or the original one. If they doesn't like the new one, i will removed it.

diznq:
While it's true that original cloak is useless, the core idea of this client is that it doesn't alter the vanilla game unless the server explicitly enables some feature. That's why all the other features are off-by-default and must be enabled by server to work. Same can be said about shaders: if server wants a different cloak shader, it should provide it's own one.

SagittaAsus:
> While it's true that original cloak is useless, the core idea of this client is that it doesn't alter the vanilla game unless the server explicitly enables some feature. That's why all the other features are off-by-default and must be enabled by server to work. Same can be said about shaders: if server wants a different cloak shader, it should provide it's own one.

The core ideea is true, but there is another feature of this. To improve things that crytek didn't. That's the main ideea of this PR. To improve cloak shader. Not everyone knows how to code. Pretty much all the servers are dead, besides of Vibing and ToxicCity.

PR #434, opened 2026-06-05
Disable ThirdPerson by default

Sagitta11:
This PR disables ThirdPerson mode. We shouldn't alter the vanilla features unless the server enables it.


diznq (approved):

diznq:
I think we can do the same for mp_hitIndicator, it wasn't part of vanilla game either :D 

Ideally it would be nice if we were able to disable third person for vehicles as well. A lot of players abuse it when they use tanks to achieve bigger turret rotation range. But I guess that would be up to a new CVar like mp_vehiclesThirdPerson with 1 being default 🙄

griseraner:
Add `mp_thirdPerson 0` to default SafeWriting config, most new servers use that. After all, every server can already easily control this, and all Mister servers disabled it for example. 
Singleplayer I think should keep it enabled, works well. As well as LAN server, I prefer not to enable it manually everytime I test something :D 
Regarding hit_indicator, I would wait until we added option to toggle it in menu. 

Vehicle thirdperson, I have a prototype fix for it in the new vehicle system, so that first and thirdperson view matches better @diznq 

mp_crymp should be default `1` though

Code/CryGame/GameCVars.cpp
```diff
@@ -604,11 +603,8 @@ void SCVars::InitCVars(IConsole* pConsole)
 	pConsole->Register("mp_flyMode", &mp_flyMode, 0, OPTIONAL_SYNC, "Enable FlyMode", OnChangeFlyMode);
 	pConsole->Register("mp_pickupObjects", &mp_pickupObjects, 0, OPTIONAL_SYNC, "Allow pickup and throw objects in DX10");
 	pConsole->Register("mp_pickupVehicles", &mp_pickupVehicles, 0, OPTIONAL_SYNC, "Allow pickup and throw vehicles (requires mp_pickupObjects 1)");
-	pConsole->Register("mp_pickupMassLimit", &mp_pickupMassLimit, 0.0f, OPTIONAL_SYNC, "Specify maximum weight possible to pickup, 0 = all objects allowed");
-	pConsole->Register("mp_pickupThrowMult", &mp_pickupThrowMult, 1.0f, OPTIONAL_SYNC, "Throw multiplier for objects");
-	pConsole->Register("mp_pickupMaxVehicleCollisions", &mp_pickupMaxVehicleCollisions, 0, OPTIONAL_SYNC, "Maximum vehicle collisions untill throwing object");
 	pConsole->Register("mp_weaponsOnBack", &mp_weaponsOnBack, 0, OPTIONAL_SYNC, "Attach weapons to back as in SP");
-	pConsole->Register("mp_thirdPerson", &mp_thirdPerson, 1, OPTIONAL_SYNC, "Allow ThirdPerson mode (F1)", OnChangeThirdPerson);
+	pConsole->Register("mp_thirdPerson", &mp_thirdPerson, 0, OPTIONAL_SYNC, "Allow ThirdPerson mode (F1)", OnChangeThirdPerson);
```
ccomrade: I just realized that changing it to zero here isn't enough. Anyone can set it to one before connecting to a server and if the server doesn't set this cvar, they will have third person view enabled.

We can do the following:
- Keep `1` as the default value.
- Set it to `0` when connecting to a server. To do this, change the value here:

https://github.com/crymp-net/crymp-client/blob/c24728eb93d6d5d73349abd410c0673a5f4ab0f3/Code/CryMP/Client/ServerConnector.cpp#L110

- Add `ServerConnector::OnDisconnect()`. Call it from `Client::OnActionEvent()` like other disconnect callbacks. Set `mp_thirdPerson` back to `1` there.

This way, `mp_thirdPerson` will be safely set to zero when connecting to a server, which is the vanilla behavior. Server owners can set it to `1` if they want to. And it stays `1` by default in singleplayer and local server.

@griseraner @diznq @Sagitta11 do you all agree with this approach?

@Sagitta11 could you please implement it here?

ccomrade (changes requested):

Code/CryGame/GameCVars.cpp
diznq: In original game there was no third person mode at all in single-player. This ain't Assassins Creed :D

griseraner: Instead of going the easy way of editing a couple numbers, maybe we should edit one of the buttons in the options menu, to toggle "prefer thirdperson to be enabled/disabled by default". It should take a fraction of the time compared to the amount of hours spent in improving the thirdperson, and it's one of the early improvements of CryMP client (just have a look at the crappy original version).
I know several people who like to use the thirdperson feature, also in SP.

diznq: Well of course they like to use it since they can gain an unfair advantage over others (also known as cheat) such as seeing behind corners, higher aim range in tanks etc... Who doesn't like that? :D For example I get pretty tired from getting killed by an AA or a gauss tank from an angle that should have been impossible if the person used first person mode 🙄.

So in my opinion it is correct if this stays as a CVar set by a server with zero being the default.

griseraner: Everybody can use it once it's enabled, so it's fair. Just like vehicle third person. It won't change much as all servers who don't want it already disabled it. 🤷‍♀️

diznq: Well, then it can be disabled by default and all servers that want it enabled can enable it :D

Anyways, personally I don't care, I care more about fixing third person in vehicles than whether third person should be enabled or disabled by default. That's the real issue anyway.

ccomrade (approved):

Code/CryGame/GameCVars.cpp
ccomrade: Looks good to me now. The only potential issue is that if someone puts `mp_ThirdPerson 0` into `autoexec.cfg` for some reason, it will be reset to `1` after disconnect, but that's fine.

PR #435, opened 2026-06-06
Update to mimalloc 3.3.2 and use it as Lua allocator

ccomrade:
Use the latest mimalloc 3.3.2 and enable it by default, but only for Lua to reduce pressure on the global allocator (`CryMalloc`) without replacing it completely. This is mainly preparation for the new sound system, which needs a fast memory allocator for FMOD.


griseraner:
Nice

griseraner (approved):

diznq:
Can't we use this globally?

ccomrade:
> Can't we use this globally?

Let's move things to mimalloc step by step to avoid risking many issues at the same time. It's possible that mimalloc itself will cause some problems that we have to fix first. It uses low-level WinAPI features such as memory mapping and thread-local storage to achieve its very high performance. This may be problematic on older Windows and/or 32-bit build. It also allocates memory on very high addresses in 64-bit, which helped discover pointer truncation bugs in the engine. All these bugs should be fixed or workarounded already, but it's still better not to rush into using mimalloc everywhere.

diznq (approved):

PR #440, opened 2026-06-25
CryPhysics Stability Improvements

griseraner:
# CryPhysics Stability Improvements

This PR adds a number of defensive fixes to improve CryPhysics stability and reduce crashes caused by invalid or inconsistent runtime state.

## Changes

* **Prevent invalid particle collision part access**

  * Added validation before accessing particle collision parts to prevent out-of-bounds access and invalid pointer dereferences.

* **Guard invalid physics change requests**

  * Added safety checks around physics state changes and queued requests to avoid invalid operations that could corrupt internal physics state.

* **Add `RayTraceEntity` safety validation**

  * Added validation for input parameters (entity, geometry pointers, hit buffer, etc.) before performing ray intersection tests.
  * Prevents crashes caused by invalid entities or malformed ray trace requests.

* **Improve typed entity list handling**

  * Improved `ChangeEntitySimClass()` by validating an entity's actual typed-list ownership instead of relying only on its previous simulation class.
  * Helps prevent duplicate list insertion and typed-list corruption, which could eventually lead to crashes during physics updates.

## Testing

These changes have been tested extensively using the CryMP pickup stress test, repeatedly picking up and dropping every physical object on the map across multiple clients over long multiplayer sessions.

During testing, the typed entity list fix successfully eliminated a rare list corruption issue that had previously resulted in physics crashes. These changes have significantly improved overall stability.

### DIFF
--- a/Code/CryPhysics/PhysicalWorld.cpp
+++ b/Code/CryPhysics/PhysicalWorld.cpp
@@ -3019,7 +3019,7 @@ void CPhysicalWorld::ChangeEntitySimClass(CPhysicalEntity* pent)
 	{
 		const int list = actualOldList;
 
-		if ((unsigned int)list >= 8u)
+		if (list < 0 || list > 7)
 		{
 			CryLogAlways(
 				"[PhysicalWorld] ChangeEntitySimClass: invalid actualOldList=%d pent=%p",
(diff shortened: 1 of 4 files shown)


Code/CryPhysics/PhysicalWorld.cpp
```diff
@@ -2924,62 +2954,157 @@ void CPhysicalWorld::DetachEntityGridThunks(CPhysicalPlaceholder* pobj)
...
+		}
+
+		return;
+	}
+
+	if (actualOldList >= 0)
 	{
+		const int list = actualOldList;
+
+		if ((unsigned int)list >= 8u)
```
ccomrade:
```suggestion
		if (list < 0 || list > 7)
```

ccomrade (approved):
Thanks for fixing these issues!

PR #441, opened 2026-06-27
Bring CrySoundSystem

ccomrade:
Here comes the sound system. Initial cleanup and refactoring are done. It loads FMOD dynamically and supports both the older C1 FMOD and the newer Warhead/Wars one. DSOAL (`-dsoal`) now works with both FMOD versions too. CrySoundSystem also needs `libvorbis` and `libogg`. The latest versions of these libraries are added into the client. Server exe doesn't contain CrySoundSystem and the libraries to keep it smaller.

Additional changes:
- Refactor Launcher and remove unnecessary `gLauncher` variable.
- Restore `BitFiddling.h` and `STLPoolAllocator.h` CryCommon headers needed by CrySoundSystem.
- Add `-oldsound` to load the original `CrySoundSystem.dll` instead for testing purposes.


diznq (approved):
Let's merge this! :D

PR #443, opened 2026-07-02
Update SSM SafeWriting

diznq:
- Add support for setting prices per team
- Update translator code and add AddTranslations for batch translation inserts
- Update translator to take language from player's locale instead of guessing it based on host-name
- Update validate command to check HWID as well
- Update IsUsable to check for in-combat status


ccomrade (approved):
Looks good!

PR #444, opened 2026-07-06
Sync with c1-launcher v8

ccomrade:
### Improved fix for 64-bit pointer truncation in CryMemoryAllocator

The main change here is clean hooking of CryMalloc functions by simply exporting them from the EXE. This works because the CryMemoryManager helper in each engine DLL searches for the functions in the EXE first and then falls back to `CrySystem.dll`. For this reason, our internal CryMalloc functions need to have the same signature as the original ones. Most of the changes are due to this.

The `STLport_Allocator` class is replaced with the improved `SafePool` class from c1-launcher. It allocates the pool as a single large chunk of memory. This allows the `Contains` function to be `O(1)` instead of `O(n)`. The pool location is chosen to avoid unnecessary relocation of engine DLLs. See https://github.com/ccomrade/c1-launcher/pull/84.

Also, `CryMemoryManagerInfo` console command is added to show the statistics.

### Abort handler in crash logger

The crash logger is now able to catch `abort()` calls.

### Additional engine patches

The following patches from c1-launcher are added:

- `CryAction::DisableGameplayStats`: Prevent CryMP-Server from automatically creating unnecessary `Game/gameplaystatsXXX.txt` files.
- `CryRenderD3D10::HookInitAPI`: Support loading custom D3D10 DLLs from Bin32 and Bin64 directories. This allows using DXVK on Windows.

### DIFF
--- a/Code/CryAction/ItemSystem.cpp
+++ b/Code/CryAction/ItemSystem.cpp
@@ -296,7 +296,8 @@ void ItemSystem::Scan(const char* folderName)
 IItemParamsNode* ItemSystem::CreateParams()
 {
 	// use CryMalloc to match CryFree used by the original ItemParamsNode destructor
-	void* mem = CryMalloc(sizeof(ItemParamsNode));
+	std::size_t allocatedSize = 0;
+	void* mem = CryMalloc(sizeof(ItemParamsNode), allocatedSize);
 	return new (mem) ItemParamsNode;
 }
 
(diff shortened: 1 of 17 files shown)


griseraner (approved):

diznq:
I would keep gameplay stats txt files, they are pretty useful when auto-generated.

diznq (approved):

ccomrade:
> I would keep gameplay stats txt files, they are pretty useful when auto-generated.

There is still the `dump_stats` console command to generate them. It works in singleplayer too.

PR #445, opened 2026-07-08
Fix ChangeEntitySimClass validation regression

griseraner:
ChangeEntitySimClass was doing a full typed-list scan on every call,
causing noticeable CryPhysics spikes.

Only fall back to FindTypedListOwner() when ownership looks wrong,
while keeping the original fast path for normal cases.

Keeps the ownership fix and removes the performance regression.


diznq (approved):

ccomrade (approved):
The lags are fixed. Well done!

PR #446, opened 2026-07-11
Add server pak support to CryMP-Server

ccomrade:
An optional server pak download link can now be set using `SafeWriting.Settings.ServerPakUrl`.


diznq:
We already have this, it's just not defined in Settings.lua, but it takes PAK = "123" from there: https://github.com/crymp-net/crymp-client/blob/e37af0f6c26c412b510c2a2198a94a8227140aa8/SSMs/SafeWriting/CryMP-Server/SafeWriting/SafeWritingMain.lua#L187

So maybe only add it to Settings.lua with nil as default, but leave Main.lua untouched :D

ccomrade:
Oops. This is what happens when rushing things. Let's fix it :D

diznq (approved):

PR #447, opened 2026-07-11
Add speed of sound

diznq:
Ladies and gentlemen, the best feature ever is here! :D

Server can optionally define speed of sound, by default it is infinite (0).
If speed of sound is defined and Distance / Speed of sound > 0.1s, then sound is delayed by given amount.

### DIFF
--- a/Code/CrySoundSystem/Sound.cpp
+++ b/Code/CrySoundSystem/Sound.cpp
@@ -14,6 +14,9 @@
 #include "Sound.h"
 #include "SoundSystem.h"
 
+#include "CryGame/Game.h"
+#include "CryGame/GameCVars.h"
+
 CSoundSystem* CSound::m_pSSys = NULL;
 
 CSound::CSound(CSoundSystem* pSSys)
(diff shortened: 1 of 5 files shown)


Code/CrySoundSystem/Sound.cpp
```diff
@@ -14,6 +14,9 @@
 #include "Sound.h"
 #include "SoundSystem.h"
 
+#include "CryGame/Game.h"
+#include "CryGame/GameCVars.h"
```
ccomrade, line 18: Hmmm including CryGame directly inside CrySoundSystem isn't nice, but it's fine for now. We need a better solution for these optionally synced cvars. Probably by extending Console and CVar sync once CrySystem and CryAction code is available.

ccomrade (approved):
Cool idea!

Code/CrySoundSystem/Sound.cpp
diznq: I would use SoundSystemCommon.cpp, but apparently Sound System is not loaded when running as server :D

PR #448, opened 2026-07-14
Add -nocrashlogger

diznq:
Add an option to disable crashlogger. 
This is useful when using --minidump option with Wine, so the program crashes properly and generates a promper dump file.

Also remove MessageBox error handler for server, since this is not very useful when server runs in a CLI :D

### DIFF
--- a/Code/Launcher/Main.cpp
+++ b/Code/Launcher/Main.cpp
@@ -26,6 +26,7 @@ int main()
 int __stdcall WinMain(void*, void*, char*, int)
 #endif
 {
+#ifdef CLIENT_LAUNCHER
 	try
 	{
 		Launcher().Run();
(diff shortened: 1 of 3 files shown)


Code/Launcher/Main.cpp
```diff
@@ -41,4 +42,7 @@ int __stdcall WinMain(void*, void*, char*, int)
 	}
 
 	return 0;
+#else
+	Launcher().Run();
```
ccomrade, line 46: Good point about message box in server. However, it's needed when running the server in Windows, which is what most people use. Also, without this `try-catch`, the server will just crash instead of showing the error message.

To have a truly console server for Linux/Wine, you can build the server exe with `CRYMP_CONSOLE_APP`. There is a CMake option for it. Although i probably forgot to add it into CMake presets :D

PR #450, opened 2026-07-15
Move ads authority from master to individual servers

diznq:
# Ads

ScreenShot0041

Instead of ads being provided by master, ads are now turned off by default unless server specifies ads resource in synched global value 3001.

The format is same as JSON from master:

```json
{
  "resources": "(empty string in this set-up)",
  "groups": [
    {
      "id": "Unique Ad Group ID",
      "viewId": "Unique Ad View ID (can be empty string in this ad serving set-up)",
      "versions": [
        {
          "id": "Unique Ad ID 1",
          "aspectRatio": 1,
          "path": "Path/To/Ad1/In/Pak.dds"
        },
        {
          "id": "Unique Ad ID 2",
          "aspectRatio": 2,
          "path": "Path/To/Ad2/In/Pak.dds"
        }
      ]
    }
  ]
}
```

## Ad group vs Ad version

As group represents ad as a whole. Imagine ad group as an ad for "SCAR is 30% off", this ad can have multiple versions with different aspect ratios. 

For example big billboard ad version has aspect ratio 1 while small road signs have aspect ratio of 2.

## Resources and View IDs

In this set-up `resources` and `viewId` fields are not used anymore. They were used in past to specify URL where PAK with ads is located and to track viewership of ads by individual players.

In current set-up the system assumes ads are already inside server pak and viewership is not tracked (perhaps in future).

## ads CVar

In past we had ads CVar that allowed players that played 25+ hours to disable ads. This is getting removed, since ads are turned off by default and are enabled by servers to serve billboards they wish, therefore it is not to be disabled if server enforces it.

### Billboard ad

Here is an image of a billboard ad (aspect ratio 1). It contains two billboards in one since game reuses same texture for two different billboard geometries just with a different crop on Y axis. Ideal resolution is 512x512. Format must be DDS.

image

### Road sign ad

Here is an image of a road sign ad (aspect ratio 2). Unlike billboard ad, it contains a single alternative. Ideal resolution is 512x256. Format must be DDS.

image

### DIFF
--- a/Code/CryMP/Client/Advertising.cpp
+++ b/Code/CryMP/Client/Advertising.cpp
@@ -33,6 +33,8 @@
 #define AD_CYCLE_ERROR 2
 
 #define ADVERTISING_SERVER_TOKEN 3000
+#define ADVERTISING_SERVER_RESPONSE 3001
+
 
 #ifndef ADS_ENABLED
 #define ADS_ENABLED true
@@ -41,6 +43,8 @@
 #ifndef ADS_SERVER
 #define ADS_SERVER "https://crymp.org/adserve/api/v1/public/ads"
 #endif
+
+// #define ADS_REMOTE
 
 struct SAdCue {
 	Vec3 pos;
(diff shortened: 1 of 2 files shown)


Code/CryMP/Client/Advertising.cpp
```diff
@@ -42,6 +44,8 @@
 #define ADS_SERVER "https://crymp.org/adserve/api/v1/public/ads"
 #endif
 
+// #define ADS_REMOTE
```
ccomrade, line 47:
```suggestion
// #define REMOTE_ADS
```
It's still only a comment though.

ccomrade (approved):

PR #458, opened 2026-07-28
Fix data race in SynchedStorage and further modernize it

ccomrade:
There is a data race bug in SynchedStorage. The root cause is CryString. It uses ref-counting instead of deep copies and cannot be shared between threads. `CSynchedStorage::SerializeValue` and `CSynchedStorage::SerializeEntityValue` are called from network thread. This causes CryString being copied between network thread and main thread, which sporadically triggers the bug.

To fix the bug, CryString in SynchedStorage is completely replaced with `std::string`. This allowed replacing CryString with `std::string` in other places as well. Additionally, `CConfigurableVariant` in SynchedStorage is replaced with modern `std::variant`.

### DIFF
--- a/Code/Cry3DEngine/TimeOfDay.cpp
+++ b/Code/Cry3DEngine/TimeOfDay.cpp
@@ -1042,18 +1042,18 @@ Vec3 TimeOfDay::CalculateSunDirection(const Vec3& sunRotation) const
 	return sunDirection;
 }
 
-void TimeOfDay::LoadCustomSettings(string xmlPath, float blendDuration)
+void TimeOfDay::LoadCustomSettings(const char* xmlPath, float blendDuration)
 {
-	if (xmlPath.empty())
+	if (!xmlPath || !*xmlPath)
 	{
 		this->RestoreLevelDefaults(blendDuration);
 		return;
 	}
 
-	XmlNodeRef node = gEnv->pSystem->LoadXmlFile(xmlPath.c_str());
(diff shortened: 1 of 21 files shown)


diznq (approved):
Very cool stuff

griseraner (approved):
👍 👍 👍

ccomrade:
Tested also on server-side:

```lua
g_gameRules.game:SetSynchedGlobalValue(32325, true)
g_gameRules.game:SetSynchedGlobalValue(32326, 3.14159)
g_gameRules.game:SetSynchedGlobalValue(32327, 42)
g_gameRules.game:SetSynchedGlobalValue(32328, SafeWriting.ChatEntity.id)
g_gameRules.game:SetSynchedGlobalValue(32329, "Test string")
```

```
-------------------------------- SynchedStorage --------------------------------
Globals:
  00000011 -      int: 0
  00000012 -      int: 0
  00000301 -      int: 0
  00000302 -      int: 0
  00001000 -   string: who knows :)
  00001002 -   string: something
  00032325 -     bool: true
  00032326 -    float: 3.141590
  00032327 -      int: 42
  00032328 - entityId: 131042
  00032329 -   string: Test string
Entity 262078 (Comrade):
  00000100 -      int: 0
  00000101 -      int: 0
  00000102 -      int: 0
  00000103 -      int: 0
  00000105 -      int: 0
  00000106 -      int: 0
  00000200 -      int: 0
  00000201 -      int: 0
  00000202 -      int: 1
--------------------------------------------------------------------------------
```

Works well. Merging now.

PR #461, opened 2026-08-01
Synchronized cloak visibility control

griseraner:
Adds `mp_cloakVisibility`, a synchronized CVar for changing the nanosuit cloak effect.

- `0` = least visible
- `1` = original Crysis
- `>1` = more visible

The shader now adjusts refraction, cloak lighting and sparks. The shared CloakLayer is updated only when the CVar changes.

Default is `1`, keeping the original behavior. Custom server cloak shaders that do not support the new parameter continue using their own effect.

ScreenShot0802
ScreenShot0801
ScreenShot0805
ScreenShot0804
ScreenShot0803

### DIFF
--- a/Code/CryGame/GameCVars.cpp
+++ b/Code/CryGame/GameCVars.cpp
@@ -632,6 +632,66 @@ void SCVars::InitCVars(IConsole* pConsole)
 	pConsole->Register("mp_strafeJump", &mp_strafeJump, 1, OPTIONAL_SYNC, "Enables or disables strafe jumping");
 	pConsole->Register("mp_fpsLimit", &mp_fpsLimit, 0, OPTIONAL_SYNC, "Sets FPS upper boundary (0: disabled)");
 	pConsole->Register("mp_soundSpeed", &mp_soundSpeed, 0.0f, OPTIONAL_SYNC, "Speed of sound (0 = infinite)");
+	pConsole->Register("mp_cloakVisibility", &mp_cloakVisibility, 1.0f, OPTIONAL_SYNC, "Controls cloak visibility(0 = least visible, 1 = original Crysis, > 1 = more visible)",
+		[](ICVar* pCVar)
+		{
+			if (!gEnv->bClient)
+				return;
+
+			IMaterialManager* pMaterialManager = gEnv->p3DEngine->GetMaterialManager();
+			if (!pMaterialManager)
+				return;
+
+			IMaterial* pDefaultLayersMaterial = pMaterialManager->GetDefaultLayersMaterial();
+			if (!pDefaultLayersMaterial)
+				return;
+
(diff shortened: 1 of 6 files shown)


diznq:
image

What if some people use their own cloak shader? Now this will overwrite it everywhere for them and they won't be able to cheat anymore 🤔

diznq:
Nice work though, I think we can merge this right away if Comrade approves :D

griseraner:
> Nice work though, I think we can merge this right away if Comrade approves :D

Thanks, one little thing I noticed the spark effect was getting a bit less when going above 1.0, could be the outdoor lights I'll have look at it tonight, but after that we can merge I guess :D

diznq:
From what I've noticed spark effect is based on outdoor lighting. If you are on sun, it is heavier than when in shadow. On Vibing I just removed that entire if/else to keep sparks gone forever :D

Code/CryGame/GameCVars.cpp
```diff
@@ -632,6 +632,66 @@ void SCVars::InitCVars(IConsole* pConsole)
 	pConsole->Register("mp_strafeJump", &mp_strafeJump, 1, OPTIONAL_SYNC, "Enables or disables strafe jumping");
 	pConsole->Register("mp_fpsLimit", &mp_fpsLimit, 0, OPTIONAL_SYNC, "Sets FPS upper boundary (0: disabled)");
 	pConsole->Register("mp_soundSpeed", &mp_soundSpeed, 0.0f, OPTIONAL_SYNC, "Speed of sound (0 = infinite)");
+	pConsole->Register("mp_cloakVisibility", &mp_cloakVisibility, 1.0f, OPTIONAL_SYNC, "Controls cloak visibility(0 = least visible, 1 = original Crysis, > 1 = more visible)",
+		[](ICVar* pCVar)
```
ccomrade, line 636: This would be nicer as a separate function instead of lambda to make the cvar list more compact, but it's just a detail.

ccomrade (approved):
Looks good. Well done!

Code/CryGame/GameCVars.cpp
griseraner: Hmm I think I prefer having the callback close to the cvar registration :D

griseraner:
> From what I've noticed spark effect is based on outdoor lighting. If you are on sun, it is heavier than when in shadow. On Vibing I just removed that entire if/else to keep sparks gone forever :D

Just checked, and the sparks don`t increase much above >1.0 which is kinda nice. Let's keep it like that. :D

MediumSolid:
Excellent work so far. But I don't know if you guys have noticed, but the shader effect also has this screen-space horizontal scanline effect. It would be nice if there was an additional cVar for controlling its visibility. Just saying.

PR #465, opened 2026-08-07
Allow importing custom weapons using server PAK technology

diznq:
This PR makes it possible to import new weapons into the game through the server PAK.

This wasn't previously possible because of network errors that stemmed from lack of new entity class registration within the network and game object contexts.

### DIFF
--- a/Code/CryMP/Common/Utilities.cpp
+++ b/Code/CryMP/Common/Utilities.cpp
@@ -47,8 +47,8 @@ void ResetGameObjectSystem() {
 	// the network system won't recognize it and it will disconnect the player on protocol error.
 	// Therefore we need to reconstruct the database like this, since this is internal function of CE SDK class
 	// and it's not exposed in the public SDK.
-	IGameObjectSystem* pGOS = static_cast<IGameObjectSystem*>(gEnv->pGame->GetIGameFramework()->GetIGameObjectSystem());
-	if (XmlNodeRef schedParams = gEnv->pSystem->LoadXmlFile("Game/Scripts/Network/EntityScheduler.xml"))
+	IGameObjectSystem* pGOS = gEnv->pGame->GetIGameFramework()->GetIGameObjectSystem();
+	if (XmlNodeRef schedParams = gEnv->pSystem->LoadXmlFile("Scripts/Network/EntityScheduler.xml"))
 	{
 		uint32 defaultPolicy = 0;
 
@@ -99,7 +99,7 @@ void ResetGameObjectSystem() {
 	// by server/client PAK
 
 	IEntityClassRegistry* pClassRegistry = gEnv->pEntitySystem->GetClassRegistry();
-	pClassRegistry->LoadClasses("LoadClasses", true);
+	pClassRegistry->LoadClasses("Entities", true);
 
 	// Merely calling Reload doesn't create new classes in the Network system, that's why
 	// Scan(...) must be called instead
--- a/Code/CryMP/Common/Utilities.cpp
+++ b/Code/CryMP/Common/Utilities.cpp
@@ -108,7 +108,5 @@ void ResetGameObjectSystem() {
 		);
 	*pDispatchSafety = false;
 #endif
-	// Merely calling Reload doesn't create new classes in the Network system, that's why
-	// Scan(...) must be called instead
 	gEnv->pGame->GetIGameFramework()->GetIItemSystem()->Reload();
 }
--- a/Code/CryMP/Common/Utilities.cpp
+++ b/Code/CryMP/Common/Utilities.cpp
@@ -5,6 +5,8 @@
 #include "CryGame/Game.h"
 #include "CrySystem/CryPak.h"
 
+extern std::uintptr_t CRYACTION_BASE;
+
 static bool StringToKey(const char* s, uint32& key)
 {
 	const size_t len = strlen(s);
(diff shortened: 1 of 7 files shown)


Code/CryMP/Common/Utilities.cpp
```diff
@@ -0,0 +1,107 @@
...
+#endif
+
+	PFNEMPLACE Emplace = (PFNEMPLACE)kMapOperator;
+
+	// By default game initializes this only once, it contains information on how to schedule
+	// entity classes on network. Now, the issue is if there are new classes in the server/client PAK
+	// the network system won't recognize it and it will disconnect the player on protocol error.
+	// Therefore we need to reconstruct the database like this, since this is internal function of CE SDK class
+	// and it's not exposed in the public SDK.
+	IGameObjectSystem* pGOS = static_cast<IGameObjectSystem*>(gEnv->pGame->GetIGameFramework()->GetIGameObjectSystem());
```
ccomrade: No cast is needed here:

```suggestion
	IGameObjectSystem* pGOS = gEnv->pGame->GetIGameFramework()->GetIGameObjectSystem();
```

Code/CryMP/Common/Utilities.cpp
```diff
@@ -0,0 +1,107 @@
...
+	if (XmlNodeRef schedParams = gEnv->pSystem->LoadXmlFile("Game/Scripts/Network/EntityScheduler.xml"))
```
ccomrade: CryPak adds `Game/` prefix to relative paths by default:

```suggestion
	if (XmlNodeRef schedParams = gEnv->pSystem->LoadXmlFile("Scripts/Network/EntityScheduler.xml"))
```

Code/CryMP/Common/Utilities.cpp
```diff
@@ -0,0 +1,107 @@
...
+			*value = p;
+#endif
+		}
+	}
+
+	// Reset entity system and item system, so they recognize possibly new classes imported
+	// by server/client PAK
+
+	IEntityClassRegistry* pClassRegistry = gEnv->pEntitySystem->GetClassRegistry();
+	pClassRegistry->LoadClasses("LoadClasses", true);
```
ccomrade: Why `LoadClasses` instead of `Entities`?


```suggestion
	pClassRegistry->LoadClasses("Entities", true);
```

Code/CryMP/Common/Utilities.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+void ResetGameObjectSystem() {
+#ifdef BUILD_64BIT
+	constexpr size_t kSchedulingParamsOffset = 0xD8;
+	uintptr_t kMapOperator = CRYACTION_BASE + 0x1B3120;
+	typedef SEntitySchedulingProfiles* (__fastcall* PFNEMPLACE)(void* map, string* key);
+#else
+	constexpr size_t kSchedulingParamsOffset = 0x6C;
+	uintptr_t kMapOperator = CRYACTION_BASE + 0x12EE00;
+	typedef SEntitySchedulingProfiles* (__fastcall* PFNEMPLACE)(void* map, void *dummyEdx, string* key);
```
ccomrade: We can use `thiscall` directly:

```cpp
struct SchedulingParamsMap
{
	SEntitySchedulingProfiles& operator[](const string& key)
	{
#ifdef BUILD_64BIT
		std::uintptr_t func = CRYACTION_BASE + 0x1B3120;
#else
		std::uintptr_t func = CRYACTION_BASE + 0x12EE00;
#endif

		return (this->*reinterpret_cast<decltype(&SchedulingParamsMap::operator[])&>(func))(key);
	}
};
```

```cpp
SchedulingParamsMap& m_schedulingParams =
#ifdef BUILD_64BIT
	*reinterpret_cast<SchedulingParamsMap*>(reinterpret_cast<std::uintptr_t>(pGOS) + 0xD8);
#else
	*reinterpret_cast<SchedulingParamsMap*>(reinterpret_cast<std::uintptr_t>(pGOS) + 0x6C);
#endif
```

```cpp
m_schedulingParams[name] = p;
```

ccomrade (commented):
Cool idea!

Code/CryMP/Common/Utilities.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+	// Reset entity system and item system, so they recognize possibly new classes imported
+	// by server/client PAK
+
+	IEntityClassRegistry* pClassRegistry = gEnv->pEntitySystem->GetClassRegistry();
+	pClassRegistry->LoadClasses("LoadClasses", true);
+
+	// Merely calling Reload doesn't create new classes in the Network system, that's why
+	// Scan(...) must be called instead
+	gEnv->pGame->GetIGameFramework()->GetIItemSystem()->Scan("Scripts/Entities/Items/XML");
```
ccomrade: What about `WeaponSystem` (ammo items)?

diznq: That's part of these XMLs already, no?

diznq: Basically, all that's needed to make stuff work is to invoke 

```cpp
CCryAction::GetCryAction()->GetIGameObjectSystem()->RegisterExtension( name, it->second.pCreator, &classDesc );
```

via `CItemSystem::ScanXML` It's not as much about ammo vs weapon as about just registering these extensions into game object system.

Code/CryMP/Common/Utilities.cpp
diznq: True, I never remember the full syntax for these thiscalls, so I usually resort to fastcall instead. But this way it's prettier yes, so I replaced it with it, thanks :D

Code/CryMP/Common/Utilities.cpp
```diff
@@ -0,0 +1,114 @@
...
+	);
+	*pDispatchSafety = false;
+#else
+	bool* pDispatchSafety = reinterpret_cast<bool*>(
+		reinterpret_cast<uintptr_t>(pGOS) + 0x28
+		);
+	*pDispatchSafety = false;
+#endif
+	// Merely calling Reload doesn't create new classes in the Network system, that's why
+	// Scan(...) must be called instead
```
ccomrade: `Reload` actually does network registration and `Scan` is no longer called here.

```suggestion
```

ccomrade (approved):
The safety hack is a bit concerning, but it looks good otherwise. Approved.

PR #467, opened 2026-08-11
Add missile approach sound

griseraner:
## Add missile approach sound

Adds approach / fly-by sounds for missiles passing close to the player.

The missile trajectory is predicted from its current speed and direction, so faster missiles can trigger the sound earlier. Guided missiles can also trigger it again after a short cooldown if they make another pass.

Also includes a couple of SoundSystem fixes:
- Added `FLAG_SOUND_NO_3D_VELOCITY` for sounds where FMOD velocity/Doppler should not be applied.
- Fixed `Set3DPosition()` passing the velocity pointer based on `pvPosition` instead of `pvVelocity`.

The feature can be enabled/disabled with `mp_missileApproachSound`. It is enabled by default, but can be overwritten by the server.

Missile approach sounds are enabled per ammo type in XML:

```xml
<missile_approach>
	<param name="sound" value="sounds/physics:bullet_whiz:missile_approach" />
	<param name="prediction" value="2.0" />
	<param name="pass_radius" value="150" />
	<param name="distance_multiplier" value="2.5" />
</missile_approach>


diznq (approved):

ccomrade (approved):
Quick review - looks good!

PR #469, opened 2026-08-16
Fix LAM and Silencer in custom weapons

diznq:
There is currently some hardcoded C++ logic for silencers and LAM modules that relies on specific string values. This change makes that logic more generic.

Without this change, LAM attachments on custom weapons would incorrectly trigger crosshair behavior, and silencer logic could break in other cases.

To address this, the logic now uses classname naming conventions: all LAM attachments must begin with `LAM`, while all silencers must end with `Silencer`.

### DIFF
--- a/Code/CryGame/Items/ItemEvents.cpp
+++ b/Code/CryGame/Items/ItemEvents.cpp
@@ -50,7 +50,13 @@ void CItem::OnEnterFirstPerson()
 
 	for (auto const& accessory : m_accessories)
 	{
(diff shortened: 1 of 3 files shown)


ccomrade (approved):
Nice

griseraner (approved):

diznq:
@griseraner in case you wanted to add custom items or weapons to your server, the trick is to add new entry in Scripts/Network/EntityScheduler.xml with type `gun`, then it must be both in server PAK loaded on server side and also in the PAK loaded on client side.

PR #471, opened 2026-08-20
Allow importing custom ammo using server PAK technology

griseraner:


diznq (approved):

Code/CryGame/Items/Weapons/WeaponSystem.cpp
```diff
@@ -573,13 +573,19 @@ bool CWeaponSystem::ScanXML(XmlNodeRef& root, const char* xmlFile)
...
 	{
 		m_pGame->GetIGameFramework()->GetIGameObjectSystem()->RegisterExtension(name, it->second, &classDesc);
+
 		pClass = gEnv->pEntitySystem->GetClassRegistry()->FindClass(name);
-		assert(pClass);
-	}
 
+		if (!pClass)
+		{
+			CryLogWarningAlways("Failed to register ammo entity class '%s' from XML '%s'! Skipping...", name, xmlFile);
```
diznq, line 585: maybe it would be better to continue the loop rather than return the entire function, it can still return false at the end if at least one of these warnings occurrs

ccomrade: True and it will actually continue because the loop is in `Scan` and it doesn't check the return value of `ScanXML`, so we are fine.

ccomrade (approved):
Looks good!

PR #472, opened 2026-08-22
Add item skins

diznq:
You know what Crysis really needs? You guessed it right: weapon skins!

We already use entity keys 1001 and 1002 for player models, so why not extend it further to weapon skins?

1001: First person skin
1002: Third person skin

Since release was postponed anyway, I think we could include this in it as well

https://github.com/user-attachments/assets/ac7e2054-9d1e-4bc4-a538-53d33184d591

Example usage

```lua
function GetWeapon(player)
	local wpnId = player.inventory:GetCurrentItemId();
	if wpnId then
		local wpn = System.GetEntity(wpnId);
		return wpn;
	end
end

local wpn = GetWeapon(player)
if wpn and wpn.class == "FY71" then
	SetSynchedEntityValue(wpn, 1001, "Objects/Weapons/Asian/FY71/fy71_khaki.mtl")
	SetSynchedEntityValue(wpn, 1002, "Objects/Weapons/Asian/FY71/fy71_khaki_tp.mtl")
end
```

### DIFF
--- a/Code/CryGame/Items/Item.cpp
+++ b/Code/CryGame/Items/Item.cpp
@@ -410,6 +410,24 @@ void CItem::Update(SEntityUpdateContext& ctx, int slot)
 		m_bPostPostSerialize = false;
 	}
 
+	int SKIN_KEY = m_stats.viewmode == eIVM_FirstPerson ? 1001 : 1002;
+	CSynchedStorage* pSSS = g_pGame->GetSynchedStorage();
+	if (pSSS) {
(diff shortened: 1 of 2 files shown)


ccomrade (approved):
Looks good!

griseraner (approved):
Looks good. Do we need this on frame code for every item? And this could benefit from a Synched storage OnValueChanged callback in future, then we could avoid On frame checks for 300+ items :)

PR #473, opened 2026-08-22
Reload textures, shaders and nanosuit material after server PAK changes

griseraner:
So that custom skins don't persist after a server change.

Include shader reload so that it's active for any server.

### DIFF
--- a/Code/CryMP/Common/ServerPAK.cpp
+++ b/Code/CryMP/Common/ServerPAK.cpp
@@ -66,18 +66,20 @@ bool ServerPAK::Unload()
 
 void ServerPAK::OnLoadingStart(ILevelInfo* pLevel)
 {
-	if (m_reloadResources)
-	{
-		CNanoSuit::ResetCachedMaterials();
+	if (!m_reloadResources)
+		return;
 
-		if (ICVar* pReloadShaders = gEnv->pConsole->GetCVar("r_ReloadShaders"))
-		{
-			if (pReloadShaders->GetIVal() == 0)
-			{
-				CryLogAlways("$3[CryMP] [ServerPAK] Reloading shaders");
-				pReloadShaders->Set(1);
-			}
-		}
+	CNanoSuit::ResetCachedMaterials();
+
+	if (ICVar* pReloadShaders = gEnv->pConsole->GetCVar("r_ReloadShaders"))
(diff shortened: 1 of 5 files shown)


diznq (approved):

Code/CryMP/Common/ServerPAK.cpp
```diff
@@ -62,6 +66,31 @@ bool ServerPAK::Unload()
...
+
+	CNanoSuit::ResetCachedMaterials();
+
+	if (ICVar* pReloadShaders = gEnv->pConsole->GetCVar("r_ReloadShaders"))
+	{
+		const int flags = pReloadShaders->GetFlags();
+
+		pReloadShaders->SetFlags((flags & ~VF_CHEAT) | VF_NOT_NET_SYNCED);
+		pReloadShaders->Set(1);
+		pReloadShaders->SetFlags(flags);
```
ccomrade, line 80: What about `ForceSet`?

```suggestion
		pReloadShaders->ForceSet("1");
```

Code/CryMP/Common/ServerPAK.cpp
```diff
@@ -51,6 +54,7 @@ bool ServerPAK::Unload()
 	{
 		CryLogAlways("$3[CryMP] [ServerPAK] Unloaded $6%s", m_path.c_str());
 		m_path.clear();
+		m_reloadResources = true;
```
ccomrade, line 57: By the way, server pak isn't unloaded in singleplayer. What about unloading it on disconnect? Not sure if it's safe to unload directly in `ServerPAK::OnDisconnect` though. It probably needs to be postponed a bit. To ensure that all disconnect callbacks are executed with the pak still loaded.

ccomrade (commented):
A few comments below, otherwise looks good!

Code/CryMP/Common/ServerPAK.cpp
griseraner: > What about `ForceSet`?

I tried but it didn't change the value (logged it 🤔)

Code/CryMP/Common/ServerPAK.cpp
ccomrade: Fixed in #475

PR #474, opened 2026-08-22
Add sound pitch control

griseraner:
Adds functional pitch control for sounds through XML and Lua :D

Item XML 

```xml
<sound ... pitch="0.8"/>
```

Lua:

```lua
Sound.SetParameterValue(soundId, "pitch", 0.8)
```

`1.0` is the default pitch; lower values make sounds deeper/slower, while higher values make them higher/faster.

### DIFF
--- a/Code/CryGame/Items/Item.h
+++ b/Code/CryGame/Items/Item.h
@@ -534,25 +534,17 @@ public:
 
 	struct SAudio
 	{
-		SAudio()
-			: isstatic(false),
-			sphere(0.0f),
-			airadius(0.0f),
-			pitch(1.0f), //CryMP
-			issynched(false)
-		{}
+		ItemString name{};
+		float airadius{};
+		float sphere{};
+		float pitch{1.0f}; //CryMP
+		bool isstatic{};
+		bool issynched{};
 
 		void GetMemoryStatistics(ICrySizer* s)
 		{
 			s->Add(name);
 		}
-
-		ItemString	name;
-		float		airadius;
-		float		sphere;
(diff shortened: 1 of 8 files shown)


diznq (approved):

Code/CryGame/Items/Item.h
```diff
@@ -534,18 +534,25 @@ class CItem :
...
-		float			    airadius;
-		float			    sphere;
-		bool			    isstatic;
-		bool          issynched;
+		ItemString	name;
+		float		airadius;
+		float		sphere;
+		float		pitch; //CryMP
+		bool		isstatic;
+		bool		issynched;
```
ccomrade: Let's remove the constructor above and do this:


```suggestion
		ItemString	name{};
		float		airadius{0.0f};
		float		sphere{0.0f};
		float		pitch{1.0f}; //CryMP
		bool		isstatic{false};
		bool		issynched{false};
```

ccomrade (approved):
`static_cast` instead of C-style casts would be nice, but it's fine. Well done!

Code/CryGame/Items/Item.h
```diff
@@ -534,18 +534,25 @@ class CItem :
...
-
-		void GetMemoryStatistics(ICrySizer * s)
+		SAudio()
+			: isstatic(false),
+			sphere(0.0f),
+			airadius(0.0f),
+			pitch(1.0f), //CryMP
+			issynched(false)
+		{}
+
```
ccomrade: See the previous suggestion.

```suggestion
```

PR #475, opened 2026-09-03
Fix reloading server PAK textures

ccomrade:
Instead of `IRenderer::EF_ReloadTextures`, which sometimes crashes DX9 renderer, simply flush the cached objects in ItemSystem and EngineCache after disconnect.

Also clean up the EngineCache implementation a bit.

And instead of unloading server PAK before connect, unload it right after disconnect. This fixes server PAK unloading in singleplayer.


griseraner (commented):
Seems to work fine from Sheffield server and switching to 7ox/something. But 7ox to something still keeps the custom SCAR skin for some reason, probably have to fix the server PAK on 7ox
ScreenShot0900   Sykes still has the modded 7ox SCAR after disconnecting from 7ox :D Looks good though :D

PR #479, opened 2026-09-15
Fix multiplayer music transitions

griseraner:
Fixes a few issues with multiplayer music.

* Fix TeamInstantAction not triggering the music FlowGraphs by using the existing TeamAction output.
* Port improved EnteredGame detection from Wars.
* Fix menu music sometimes continuing or restarting after a level has finished loading.
* Clear the previous music theme on level load, fixing maps without music FlowGraphs inheriting music from the previous map.
* Prevent the delayed menu music timer from starting the menu music again after entering a level.


ccomrade (approved):
Looks good

Reddbr34:
Does this also fix the issue where the same track would play infinitely on a instant action match? like it never swaps to another track.

griseraner:
Not so sure about that, it changed tracks for me when I was testing IA, but I've noticed that before too.. You can test the dev build, let me know if it still happens there

